PR #1, opened 2019-01-19
Update user_io

gyurco:
I've noticed you're using a very old user_io. Here's an updated one, used by all my updated cores now, without clock domain crossing issues.
I can add 15kHz and YPbPr support, too.
I think rgb and vga clock domains should be put to a asynchronous clock groups (or I would simpy use the rgb clock for vga, 1 MHz difference would work on most monitors, and don't worth the extra hassle with async clocks).
Also added a small data_io change to get rid of another clock domain crossing.


jotego:
Thanks a lot! I will check the files now. Can you try running the core with
the attached rom? Apparently I provided a bad one in the dropbox link.
Shame on me. You should get the game running with it.

gyurco:
It works! Excellent core.
I like the TV-like filter, looks very authentic. Will also append YPbPr support, it's easy. 15 kHz needs a new HSync generator, can do it later also.

PR #2, opened 2019-01-20
Add 15kHz TV mode

gyurco:
Had to update the OSD, too, it didn't work well on TV (and now it has only 1 clock).


jotego:
Wow! Great! Thanks :-)

jotego:
@gyurco apparently the 15kHz mode does not work with the latest release. The game clock (clk_rgb) was 24MHz when you added this function. Now I have been able to push it down to 12MHz. I have not modified you code for 15kHz signal (at least not intentionally). Do you think it needs 24MHz to work?

PR #12, opened 2019-03-03
MiST: OSD with rotation

gyurco:
No need for firmware modification.
In 1942, the screen flip is immediate with the OSD, but delayed to reset with the game. I didn't want to touch the inner logic, I think you can easily fix this.

Comment: screen filter is exchanged in gng (again).


jotego:
This is great! Thanks a lot :-)

PR #43, opened 2020-03-04
Align VSync to HSync

paulb-nl:
VSync should be aligned to the rising edge of HSync

This fixes Direct video


jotego:
Thanks, thanks so much! I would have thought the problem came from an interaction with the MiSTer framework.

PR #57, opened 2020-09-04
Biocom: adjust CEN's to the MCU's needs

gyurco:
Controls work!

### DIFF
--- a/cores/biocom/hdl/jtbiocom.def
+++ b/cores/biocom/hdl/jtbiocom.def
@@ -10,7 +10,7 @@ VIDEO_HEIGHT=224
 
 [mist|sidi]
 # Sound doesn't fit in small devices
-NOSOUND
+#NOSOUND
 OSD_NOBCK
 VIDEO_WIDTH=384
 JTFRAME_SCAN2X
--- a/cores/biocom/hdl/jtbiocom_main.v
+++ b/cores/biocom/hdl/jtbiocom_main.v
@@ -253,12 +253,13 @@ end
 
 /////////////////////////////////////////////////////
 // MCU DMA data output mux
-always @(*) begin
+always @(posedge clk) begin
+	if (mcu_cen)
     case( {mcu_obj_cs, mcu_ram_cs, mcu_io_cs } )
-        3'b100:  mcu_din = oram_dout[7:0];
-        3'b010:  mcu_din = wram_dout[7:0];
-        3'b001:  mcu_din = cabinet_input[7:0];
-        default: mcu_din = 8'hff;
+        3'b100:  mcu_din <= oram_dout[7:0];
+        3'b010:  mcu_din <= wram_dout[7:0];
+        3'b001:  mcu_din <= cabinet_input[7:0];
+        default: mcu_din <= 8'hff;
     endcase
 end
 
@@ -272,7 +273,7 @@ always @(*) begin
         // MCU access
         work_A   = mcu_addr[13:1];
         work_uwe = 1'b0;
-        work_lwe = mcu_wr ;
+        work_lwe = mcu_wr;
         ram_cen  = mcu_cen;
     end else begin 
         // CPU access
@@ -283,7 +284,7 @@ always @(*) begin
     end
 end
 
-jtframe_ram #(.aw(13),.cen_rd(1)) u_ramu(
+jtframe_ram #(.aw(13),.cen_rd(0)) u_ramu(
     .clk        ( clk              ),
     .cen        ( ram_cen          ),
     .addr       ( work_A           ),
@@ -292,7 +293,7 @@ jtframe_ram #(.aw(13),.cen_rd(1)) u_ramu(
     .q          ( wram_dout[15:8]  )
 );
 
-jtframe_ram #(.aw(13),.cen_rd(1)) u_raml(
+jtframe_ram #(.aw(13),.cen_rd(0)) u_raml(
     .clk        ( clk              ),
     .cen        ( ram_cen          ),
     .addr       ( work_A           ),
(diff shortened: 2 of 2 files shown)


gyurco:
Sound fits on MiST, but dies very soon.

jotego:
Thanks a lot for that! I'm eager to test it.

Broken sound may get fixed using the latest JTFRAME, if that doesn't fix it we're back to the mastery. I'll check it. 

Maybe the sound fits now in MiST after the MiSTer scandoubler stopped being the default for MiST. That modules takes a lot of BRAM because of the HQ2X interpolator, which most people don't like anyway.

gyurco:
Unfortunately the latest JTFRAME didn't help. Debugging is hard on MiST, because now there's really no more BRAM left for SignalTap. Maybe disabling jt51, but the timers should allow at least PCM sounds?

gyurco:
Looks like there's a simple solution: as WAIT_n is broken in T80, it's best to not use it (T80pa did some corrections, but still not 100% perfect without some help from outside). But using:
```
    .CEN        ( cen & wait_n ),
    .WAIT_n     ( 1'b1        ),

instead of the wait signal seems to work

**Comment by jotego** [comment]
I've looked many times at the WAITn implementation in the T80 core and couldn't find why it fails. True that not understanding the whole state machine is not the best place to debug from...

I think I have tried that before and got intermmitent results. Let me try again.

**Comment by jotego** [comment]
It didn't work for me. I have commited the change in JTFRAME that makes that arrangement via a parameter. Have a look: https://github.com/jotego/jtframe/commit/c849c513fcb568411a5cf51d7a5d08d43329a6b7

**Comment by gyurco** [comment]
Probably it latches wrong DI data during wait in some cases (contrary to the MCU IP, which looks like doesn't latch DI at all during the memory cycle, that's why it needs it one cycle _after_ the actual memory address setup) .
T80pa totally manages WAIT_n on its own, without using T80.vhd, that's why it works (and battle-tested in Amstrad CPC). However as cycle-perfection not required here, and wait is not used in the original design, it doesn't matter.

It might be a bit better:
```
    .CEN        ( cen & (wait_n | mreq_n) ),
```
if the ROM request can be started right after the address is visible.

**Comment by gyurco** [comment]
> It didn't work for me. I have commited the change in JTFRAME that makes that arrangement via a parameter. Have a look: [jotego/jtframe@c849c51](https://github.com/jotego/jtframe/commit/c849c513fcb568411a5cf51d7a5d08d43329a6b7)

It works for me, if I set USEWAITN to 0 (this has to be the default, or at least one should be able to set this outside of JTFRAME)

**Comment by jotego** [comment]
I am trying it on the sf branch (jt_gng2 repo) and you may be trying it on the stable branch (jt_gng repo). SF uses the same jtbiocom_sound module. It didn't produce any sound either but by latching the signals -and God knows what else- it worked for SF consistently.
It doesn't produce sound for biocom now. Let me try compiling again with stable version of jtbiocom_sound and see it that sounds.
I still think that the compiler gets something wrong because it seems to be sensible to stuff that shouldn't matter. And I have never got the sound to fail in simulation.

**Comment by gyurco** [comment]
I'm on f9f6a294 in jt_gng
on f27963d in jtframe

The only local changes: 
```
-parameter USEWAITN=1;
+parameter USEWAITN=0;
```
in jtframe

and

```
-    .cen        ( cen_alt     ),
+    .cen        ( cen_fm      ),
```
in jtbiocom_sound.v

gyurco:
I found interesting how broken the 68k sound latch design in the original HW, its output enable is active only during active NMI signal, but it's too short to be useful. Then they hacked in the sound latch via the MCU.

jotego:
The original file from stable doesn't work either. I do think there is something wrong with the Quartus compiler. You can say I am in denial. Haha

The interrupt logic may be involved too as I only see this problem when JT51 and Z80 operate at the same odd clock enable rate of 3.57MHz and the Z80 is being interrupted by the JT51.

The other case when Z80 has been problematic is DD2 as you know.

I should try compiling for a different platform, Xilinx, and see if there is sound there but I'm only human and tired of this issue.

gyurco:
I'm using the good old Quartus 13.1.4 for MiST. Maybe you're using newer ones for SiDi/MiSTer?
I can send my rbf for MiST, I wonder if my device is somewhat special :)

jotego:
Your two earlier comments justs poped up now.

I think PCB design relied on the parasitic capacitance of traces for these things. Signals would have lasted a bit longer due to that. I read on an interview to CAPCOM designers of that era that they had a lot of reliability problems. No wonder why.

It's fair to say that they were pushing that logic a lot with those massive boards at high frequencies.

I am using Quartus 13.1.0. I should try your version. Let me try and find it. I don't think it's MiST. I think it's Quartus. I have seen Quartus remove items in the schematics view when these errors occur without warnings in the logs.

jotego:
And sound coming and going aways should call for a timing issue, but there are not timing problems. Latching signals (the interrupt-related signals) should fix timing. But that didn't work either.
It's a hell of a mistery.

jotego:
No 13.1.4 for linux. But I get the same errors on MiSTer version, though. Maybe I should give a go to ZX-DOS and Xilinx. The problem is they don't have a proper firmware to download the ROM. That's why I have never developed to ZX-UNO/ZX-DOS.

gyurco:
I'm using 13.1.4 on Linux. There's an Updates tab on the download page. It should be installed on an existing 13.1.0.
Sent my rbf (+ROM and ARC files just in case).

gyurco:
> I think PCB design relied on the parasitic capacitance of traces for these things. Signals would have lasted a bit longer due to that. I read on an interview to CAPCOM designers of that era that they had a lot of reliability problems. No wonder why.

No way the NMI signal lasted until the NMI handler actually called. It's active only for one 68k memory cycle (as the decoder enabled by io_cs). I wonder if the Z80 can catch it in every case (but as the sound code doesn't use IN/OUT commands - the longest on the Z80 - probably it's just OK).

PR #58, opened 2020-09-05
Biocom: fix sound + update jtframe

gyurco:
- use the same clock for the CPU as for FM
- don't use WAIT_N in the CPU
- use the CPU in JTFRAME


gyurco:
Well, works for me. I have no idea what should be the problem in your setup.

jotego:
The RBF you sent me works. Let me try compiling this one locally next.

Issue #209, opened 2023-03-02
sch: Verilog for x1_007


jotego:
Furrtek has extracted [the logic for the X1-007](https://github.com/jotego/jtcores/blob/06683beded0347228bb26bddb3399bb3d3df21a2/cores/kiwi/doc/x1-007.v). I see it in The NewZealand Story. @skutis could you add it to the schematics? It's quite short. We don't need the port definition, just the code in the middle that shows the functionality.

I'm not sure whether it's used elsewhere. @etriana85 can you check whether it is on Extermination or Insector X? You could add a link to the symbol to https://github.com/jotego/jtcores/blob/06683beded0347228bb26bddb3399bb3d3df21a2/cores/kiwi/doc/x1-007.v as the datasheet

etriana85:
verified and added.
![#202](https://user-images.githubusercontent.com/92724765/222364712-016064e0-8c0c-48f2-a429-04b02d1036ca.png)

PR #210, opened 2023-03-02
Added verilog code for X1-007

skutis:
Added verilog code for X1-007. Issue #209 .


jotego:
This PR has conflicts in tnzs files. We can solve them by overwritting the files but I wonder how did the conflicts arise if you are the only one editing those files. Maybe you are editing the files in two different folders or repositories?

Typically these conflicts arise when you a file has been modified by two different people and the same lines have been touched by the two developers. But if a single person can cause the problem if he is editing the file from two locations. No big deal, but I just wanted to make you aware of it.

PR #228, opened 2023-03-25
ffight: add third button

Abadede:
Although not mentioned in the manual, the third button allows to escape grapples.


jotego:
Does that really work? Ok, but I had never heard of it.

Abadede:
Yes it is, you can even interrupt Andore's flying pile driver mid-air.

PR #229, opened 2023-03-25
cps1: set default DIP switches

Abadede:
Settings copied from manuals.


jotego:
You're setting 32 bits in some cases. There are only 24 bits for settings in this system.

Abadede:
@jotego My bad, wasn't sure about those last two digits. I left them in because that's how it's written in the mra files for games that have a service switch. Anyway they're removed now, I hope it works for you.

jotego:
Have you tried these out? Are the games starting with the conditions you wanted?

Abadede:
@jotego Yes I think I checked everything. Off the top of my head the biggest differences with MAME defaults are:
- factory settings for some games are supposed to be demo sound=off
- daimakai, striderj, willowj default difficulty is 4 not 2 and cabinet is 2 players table
- qad and qadjr have different DIP switches.

The only thing I am not quite sure of is Strider undocumented "difficulty on life loss" switches. According to MAME source values for these are -1 (on on), Default (on off, is it -2?) and -3 (off off). USA manual mentions the automatic difficulty drop on life loss is -1. As far as I know Japanese manual doesn't tell anything. I don't understand how operators were supposed to guess that these switches should be set to on off.

jotego:
Maybe if you into the service mode there will be a sentence about the DIP settings in Strider. I've seen them quite often and they react to the changes you do in the OSD.

PR #271, opened 2023-06-05
daimakair: add starting level dipsw

Abadede:
dipsw 16,17 in daimakair not only control the number of lives but most importantly the starting level (with dipsw 18,19). I didn't add the number of lives (ie. level 1 starts with 3 lives, level 1-2 with 4, 2 with 5, etc.) but I can if needed.

### DIFF
--- a/cores/cps1/cfg/mame2mra.toml
+++ b/cores/cps1/cfg/mame2mra.toml
@@ -88,6 +88,7 @@ rename=[
 ]
 extra=[
     { machine="pang3", name="Freeze", options="On,Off", bits="23" },
+    { setname="daimakair", name="Starting Level", options="-,-,-,-,6,6,5-2,5,4-2,4,3-2,3,2-2,2,1-2,1", bits="16,19" },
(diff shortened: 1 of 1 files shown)


jotego:
Did you try running 
`jtframe mra cps1`
?

jotego:
The problem with these DIP switches is that they hold different meaning depending on whether the test bit is set or not. We do not have a framework for this functionality at the moment. I am not even sure whether MiSTer can have the text for the DIPsw menu changed on the fly.

We cannot have an MRA file that supports both sets of DIP switch values: regular ones and test mode ones.

I am sorry but we cannot merge this request.

Abadede:
Sorry for the late reply. In daimakair level select switches don't depend on the test bit.
As for ghoulsu, test switches could just be added under 'Game Mode' in the menu. Changing test switches would of course also alter the corresponding normal game switches and one would have to move them back to their liking when going back to normal game mode, but that's how it is on a real board anyway.

PR #456, opened 2023-12-13
Scandoubler fix and handle blanks

gyurco:
- wraddr is incremented with alt_pix, so reset it with alt_pix, too. This fixes cores, like 1942.
- handle blank signals, and emit a scandoubled DE for HDMI usage


gyurco:
Some cores are broken on MiST, as every other line has a different length (depending on the HSYNC position, as it's sampled at the wrong time), the picture is "dancing" on the VGA screen. Also the OSD is sliced in two halfs. This PR fixes it.

For HDMI usage, a good DE signal is needed, so use the original blanks to create a scandoubled DE.

jotego:
Thank you

PR #472, opened 2024-01-02
Add new demistified targets (zxtres, deca, de10lite) and mist compatible target psd_gx150

somhi:


gyurco:
It's not my repo, but I think you should add the DeMistify framework as a submodule.

somhi:
That would be ideal, of course, but I did not want to mess too much outside of the target/demist_ folders.   
When I did it, I was not sure how to link the modules from a DeMiSTify submodule in my target.yaml and working with a submodule not part of Jotego submodules complicated things for me.
Maybe in a second revision I could do that with some guidance of José.

somhi:
Sorry I have to close this PR, as it's getting many commits and now has conflicts with upstream repo. Moves to #484

PR #484, opened 2024-01-09
Add new demistified targets (zxtres, deca, de10lite) and mist compatible target psd_gx150

somhi:
Clean and no conflict PR as was #472


jotego:
Are the time violations correctly checked for these new targets? Particularly the non-Quartus targets will produce the warnings in a format that jtcore does not catch and may be missed.

somhi:
No timing violations for targets demist_de10lite and mist_psd_gx150
demist_zx3* and demist_deca have time violations. 

I could have a try with the _deca target to see if I can clear something but I'm still not understanding fully how to constraint the cores.  There are also warnings I could clear up.

_zx3* target  I'm opening generated project in Vivado and there are timing violations.   I'm still quite new to Xilinx and Vivado so it  will take some time before I can try to clear those violations.

I also see some improvement in the demist_ and mist_ targets in order to avoid recurrent maintenance.  Those targets have a modified version of some mist files (jtframe_mist.sv,  jtframe_mist_base.sv and jtframe_mist_top.sv). Every time those files have changes, I need to update also for those targets.    I can modify those files from mist target with Verilog macros so they are the same for mist, mist_* and demist_* targets.

Maybe it's better to close the PR, do more work on my side, and do new PR later.  But first please, let me know if the way I did the PR in those three commits is Ok for you, or is better to generate separate PRs, ....

cores/aliens/cfg/macros.def
```diff
@@ -31,7 +31,7 @@ JTFRAME_DIPBASE=8
 # JTFRAME_IOCTL_RD=19472
 JTFRAME_INPUT_RECORD
 
-[mist|sidi]
+[mist|sidi|demist_*|mist_*]
```
jotego, line 34: I think `[*mist*|sidi]` would work too

cores/outrun/cfg/macros.def
```diff
@@ -66,6 +66,14 @@ JTFRAME_CREDITS
 JTFRAME_CREDITS_PAGES=3
 JTFRAME_CREDITS_HSTART=32
 
+[demist_zx3a35]
+-JTFRAME_SKIP
+JTFRAME_LF_BUFFER
+JTFRAME_OSD_NOLOGO
```
jotego, line 72: No BRAM even for the logo?

modules/jtframe/bin/jtcore
```diff
@@ -523,61 +544,65 @@ if [ $SKIP_COMPILE = FALSE ]; then
...
-        quartus_sh --flow compile $corename > "$LOG"
-        if ! grep "Full Compilation was successful" "$LOG"; then
-            grep -i error "$LOG" -A 2
-            echo -e "${rBG}ERROR:${NC}${wBG} while compiling the project. ${red}Aborting${NC}"
-            report FAIL
-            exit 1
-        fi
-        if [ ! -e $corename.map.rpt ]; then
-            RPTPATH=$OUTPUTDIR
+        if [[ $TARGET = demist_zx3a* ]]; then
```
jotego, line 547: we are indenting a lot of code for basically not doing anything in demist_zx3a, maybe a refactor here would help to make this more readable

modules/jtframe/target/demist_de10lite/demistify/de10lite/constraints.sdc
jotego, line 1: why is this in two subfolders? We should limit the depth when possible

modules/jtframe/target/demist_de10lite/jtframe_de10lite_top.vhd
jotego, line 1: No VHD files allowed. They cannot be merged in simulation with verilog files

jotego:
The PR adds many targets, which is not necessarily good. I am worried of adding support for a device with two users.

somhi:
Ok thanks for comments.  I will rethink it before attempting new PR.

cores/aliens/cfg/macros.def
somhi: I think *mist* is not good as it will include mister as well.

cores/outrun/cfg/macros.def
somhi: I'll check if it has. I suppose yes, but at first try I disabled everything.

jotego:
Thank you for your continuous effort. It is a tremendous accomplishment.

modules/jtframe/target/demist_de10lite/demistify/de10lite/constraints.sdc
somhi: It was just to indicate those files that are part of DeMiSTify.  
What do you think of including DeMiSTify as a submodule and where should be located ?
https://github.com/robinsonb5/DeMiSTify

modules/jtframe/target/demist_de10lite/jtframe_de10lite_top.vhd
somhi: No problem, top level can be changed to Verilog.

jotego: Exactly. I did not mean all of them, but at least the jtframe ones. Open source tools cannot run sims with merged languages.

modules/jtframe/target/demist_de10lite/demistify/de10lite/constraints.sdc
jotego: I don't understand the value of DeMiSTify. It seems to be meant for MiST cores, but JTCORES already have a framework. Why not support the target FPGA directly?

cores/aliens/cfg/macros.def
jotego: Right!

modules/jtframe/target/demist_de10lite/demistify/de10lite/constraints.sdc
somhi: DeMiSTify is a great resource to support porting MiST cores to those boards without an external processor or microcontroller.  
So basically DeMiSTify generates a minimal MiST firmware to be fitted in BRAM and adds a softCPU [controller](https://github.com/robinsonb5/EightThirtyTwo) into the core to be ported that manages everything as does the SAMD controller of the MIST (OSD, SD card access, keyboard & mouse controls, and passes everything to the FPGA core).

Issue #493, opened 2024-01-18
contra,comsc: broken in 4c0d815ae745ce10cc552f8fbacc8f9d08714e48


jotego:
These two cores are not working in 4c0d815ae745ce10cc552f8fbacc8f9d08714e48. The problem must have happened much earlier.

gyurco:
Looks similar to the Karnov one...
Here (1001) the CPU should read "01", not "0F":
![image](https://github.com/jotego/jtcores/assets/8644936/ef59dc4b-bfde-4073-9807-e61f36c03fd0)

AFAIK RAM is in BRAM here, so no SDRAM controller is involved.

gyurco:
More possible that CEN rate again...the read of '1001' is only 2 master cycles. That would be a 24MHz 6809 (or 48?).

gyurco:
But CEN is usually good, only some reads are speed up...which makes the 'clock recovery' suspicious. I think my previous report about the flawed ram_ok is still stands, the input data is delayed by one cycle to ram_ok, so the CPU reads it too early.

gyurco:
Uhh, is it the intented CE?
It's similar to the driver who push full gas, then full break, and so on...
![image](https://github.com/jotego/jtcores/assets/8644936/87c0eeaa-9e2e-44c4-a678-3bc3d7b5ff2c)

gyurco:
I don't thing cen3 delaying does anything - the cpu E and Q enable is generated in jtframe_sys6809.v using cen3 and rom_ok. Something must be done in that module if cen delaying is the way to go - but personally I think it would be ugly and could have many unwanted side-effects to other working cores, I would not touch it.

gyurco:
Additional to the above, CEN delaying doesn't work on RAM in BRAM access - as it doesn't have any OK signal, so if the clock recovery thinks it should run in full speed, the same problem occurs.

jotego:
Full speed recovery is still a 50% duty cycle, so the BRAM will work. That is by design.

Delay the rom_ok by one clock cycle. That should work. Then, we can keep registering cpu_din.

Also, note that *mx5k* may be failing too because cpu_din is also registered in that core and it shares the same _jtcontra_main_

gyurco:
_Full speed recovery is still a 50% duty cycle, so the BRAM will work._

...but...it doesn't :) Look at the first waveform. It's during the RAM test code, it fails very early.

[...]

PR #494, opened 2024-01-18
Contra: make cpu_din combinatorial (#493)

gyurco:
Delayed cpu_din misses the CEN, making the CPU read wrong data

### DIFF
--- a/cores/contra/hdl/jtcontra_main_decoder.v
+++ b/cores/contra/hdl/jtcontra_main_decoder.v
@@ -75,16 +75,16 @@ always @(*) begin // Decoder 007766 takes as inputs A[15:10] and A[6:5]
     out_cs      = A[15:10] == 6'b0000_00 && A[6:4]==1 && !RnW;  // 18-1F
 end
 
-always @(posedge clk) begin
+always @(*) begin
     case(1'b1)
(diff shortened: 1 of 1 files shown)


jotego:
Try delaying cen instead of making the logic combinational

jotego:
Also, the Verilator linter test must pass for PR's

jotego:
Make the PR directly **from a branch in this repository** (after you accept the invitation) so the actions run correctly.

gyurco:
Actually I didn't get any timings worse than before.

jotego:
> The linter failed, because of: Error: Input required and not supplied: token

It is because of not having the right token. I think that if you make the PR from a branch in this repo, the actions will work.

You can run the linter locally too with `$JTFRAME/devops/lint-all.sh`. I normally run it from the $CORES folder. It leaves a lot of .log files that you can delete.

gyurco:
Running from my account fails because of the non-public submodules. But doing inside jtcores seems to work. Maybe you can set up protection to the master branch to prevent accidental pushing into it.

gyurco:
Additional to the above, CEN delaying doesn't work on RAM in BRAM access - as it doesn't have any OK signal, so if the clock recovery thinks it should run in full speed, the same problem occurs.

PR #496, opened 2024-01-19
Contra: don't delay cpu_din for Contra and variants

gyurco:
MX5K need the port_in as combinatorial, too, otherwise controls won't work.


gyurco:
Fixes #493

Issue #174, opened 2022-01-22
s16: MCU (i8751) games do not boot


sqwirral:
`jts16_20211105.rbf` works fine, but I couldn't get any builds newer that to load the game.

Body Slam (World, S16) [8751 317-0015].mra

jotego:
The MCU was not sampling the data from the main bus correctly. Yet fixing it in 8f91537 made no difference.

gyurco:
In Body Slam, there's a bus request-bus grant cycle, after that the MCU resets the CPU and never releases the reset. So it might read something scary during bus arbitration?

gyurco:
Looks like it's not deterministic, it can reach CPU reset in various ways. Maybe because of the non-cycle exact MCU.

jotego:
One thing about the SEGA System 16 is that RAM had a battery so the game usually expects the RAM to have some valid content. It can initialize it too, but you often find that it reacts to old data. That might be affecting it. Maybe it is deterministic after a power cycle (power off and on).

It might be possible to wipeout the RAM (held in the SDRAM) by modifying the game load process, but I wouldn't go into that rabbit hole just yet. Probably a power cycle will do.

The MCU is meant to get access to the main CPU bus and probably check stuff, particularly check that the ROM has not been tampered with. That was my main focus when debugging it.

gyurco:
Strangely If I reload the core via USB Blaster (clearing registers and probably BRAMs), then the game starts.

I observed the MCU does the coundown from 3 to 0 at the beginning (it can write to VRAM), but currently only 0 flashes for a moment after a reset, then it hangs.

gyurco:
Some observations:
- MCU to CPU reset is inverted (p1[6])
- MCU doesn't use bus arbitration for 68k bus access, but put the 68000 into reset while accessing RAMs/ROMs

gyurco:
- Something must be done with the PPI, too (maybe use cpu_rst), as during initialization, when the CPU mostly in reset, the video must be enabled. But Using the cpu_rst for PPI also activates flip, too, which is unneeded.

Implementing the 3 observations above makes the countdown 3-2-1-0 appear (just flipped). It still freezes after that, debugging continues...

jotego:
The reset signal can be a bit tricky too. I think it was in the JTS16B where a bug was fixed by adding support to the M68000 driving its own reset pin low. Not sure if that could play a role here too.

gyurco:
[...]

PR #523, opened 2024-01-28
S16mcu fixes (#174)

gyurco:
Fix i8751 games, add Quartet controls.

Still an issue: boot screen is flipped. Maybe i8255's reset values should be changed?


jotego:
Thank you. It looks good overall but the complexity of resetting the MCU memory should be part of the MCU module and not handled over to the core. So the _jtframe_8751mcu_ port interfacte can be kept constant. That way the other cores using this won't see the change. Also the idea of all these wrappers is to make design easy, and passing the reset task a level above goes against that principle. I suggest these changes:

- keep the old _jtframe_8751mcu_ port interface
- make a new _jtframe_ram_rst_ module, with the _jtframe_ram_ interface plus a _rst_ pin
- include the reset logic inside that module
- restore the files for other cores (biocom, btiger,cop,karnov,s16b,sf) 

That way, changes to all modules, including 8751mcu itself, are minimized. The new _jtfame_ram_rst_ module will be useful in other cases and easy to swap with the regular _jtframe_ram_ module. This will also automatically add the functionality to the other cores, which could be sensitive to this issue too.

Thank you, it is going to be a good new feature for the MCU

gyurco:
IMHO the current method is more versatile, as you can fill with arbitrary data, or just inject one byte to RAM.

But there's a bigger concern: if I put the reset logic into the MCU, I cannot control when the external reset is released, and that would be a problem if the MCU should be run, but the RAM is still in clearing phase.

jotego:
The reset is held during all the ROM download time, which is very long in comparison to the RAM length. There is a minimum time for user-triggered resets too. I will check how long it is.

I see your point about versatility but I think the right tradeoff is for game-module simplicity.

gyurco:
Here's more important to reset the RAM when the 'reset' option is chosen in OSD (or by button). When loading the core, the BRAM is initialized to 0 anyway.

jotego:
> Here's more important to reset the RAM when the 'reset' option is chosen in OSD (or by button). When loading the core, the BRAM is initialized to 0 anyway.

Yes, jtframe_board controls the length of that reset

jotego:
Another option is to make the 128 bytes available from non-BRAM memory, so reset acts on a single clock cycle. The original memory may have been like that. In that case, there is no need for a reset address counter.

gyurco:
That's 1024 registers  + other logic, I think it's too much.

Adding automatic BRAM reset would need to test all other cores which use it, sorry, but I'm a bit reluctant to do it. The current reset logic is only +4 lines, and works well so far.

jotego:
I'll change it myself then.

gyurco:
> I'll change it myself then.

Your repo, your choice.
I just genuinely think that:
- It causes more problems than it solves.
- Only one core required it so far, thus there's no benefit yet from the generalization.

jotego:
No problem. Thanks for fixing the bug!

Issue #525, opened 2024-01-30
s16: SDI graphic issues


jotego:
> However I noticed on stage 2, there are a lot of blank lines at the right edge of the screen when lot of objects attacking.
> Also looks like the scrolling is broken at specific places, probably a counter is overflowing at the wrong time.
I can look at these if you want.

The old jts16 repository (archived now) contained mentions to these problems but no clues about why they happened. So I start a new fresh issue, with @gyurco's comments in #174

jotego:
Thank you for the PR!

PR #527, opened 2024-01-30
S16: scroll layer fixes (#525)

gyurco:
- Prevent an unexpected overlow on eff_hscr
- Use pxl_cen2 for linebuffer writes

Fixes SDI, no problem with Cotton title&attract mode. Also checked Golden Axe, still looks good.


gyurco:
I decreased the delay for scr_ok by one cycle, and also my captures shown that there are plenty of time left for the next line to start, so I'm not worried.

Issue #181, opened 2022-12-12
kiwi: support for triple Z80 games


jotego:
Kabuki requires three Z80 CPUs, following the [schematics for tnzs](https://github.com/jotego/jtbin/blob/master/sch/tnzs.pdf). Currently the **JTKIWI** core is designed for the [dual Z80 version](https://github.com/jotego/jtbin/blob/master/sch/arknoid2.pdf) of the board. Resource usage report in SiDi is:

> Total logic elements : 12,239 / 22,320 ( 55 % )
> Total memory bits : 500,478 / 608,256 ( 82 % )

so adding a third Z80 to the mix should be fine. There are four games identified as triple Z80 (see kiwi/cfg/mame2mra.toml).

Adding support for triple Z80 games requires:

- [x] SDRAM mapping for third Z80 (mame2mra.toml and mem.yaml)
- [x] Connect the third Z80
- [x] Review other connections in case something else is different
- [x] Simulation starts up correctly (check both tnzs and kabuki)
- [x] Music sounds
- [x] Synthesis ok
- [x] Playable (tnzs, kabuki)
- [x] Button names in mame2mra.toml for _kabuki Z_

jotego:
Kabuki requires three Z80 CPUs, currently the core is done with only two

gyurco:
Added inputs and palette RAM handling to KabukiZ, but seems even the graphics HW needs adjustment, both tilemaps and sprites are messed up.

gyurco:
Sound added to KabukiZ. For TNZS, it needs some adjusments, probably will require another flag in mame2mra.toml, but not a big deal.
Graphics however is garbage.

gyurco:
Another problem the core no longer fits for MiST/SiDi. Maybe Kabuki can be skipped for these or the big banked RAM could be moved into SDRAM.

jotego:
I tried compiling it. It's awesome that it's booting up, colors are ok, inputs work... very good progress.

I think we can skip this for MiST/SiDi with some precompiler setting, like the **NOMCU** used in other cores. I'd rather not compromise the core performance or complexity.

The graphics seem to have some bad address mapping in part of the demo. The schematics may give hints about that. The most worrying part are the sprite glitches, which may be something else. Hopefully, it only requires applying some bit transformation to the tile and sprite ROM addresses based on some bank register or similar.

gyurco:
What's the difference between 'setnames' and 'machines' in the .toml file? Both are used in the data= section.

[...]

PR #563, opened 2024-03-01
Kabukiz (#181)

gyurco:


gyurco:
Kabuki-Z sound board and video DMA implementation for #181

Definitely needs more regression testing, but everything worked for me so far.

cores/kiwi/hdl/jtkiwi_gfx.v
```diff
@@ -224,16 +224,62 @@ jtkiwi_obj u_obj(
 // memory for the CPU
 // In MAME the lower half is called spritecodelow
 // and the upper spritecodehigh
+
+reg  [9:0] copy_addr;
```
jotego, line 228: I normally would have called this dma_ instead of copy_. But it's fine. No need to change it.

jotego:
Looks awesome. I am going to merge it and test it straight away :-)

cores/kiwi/hdl/jtkiwi_gfx.v
gyurco: I thought about it, but it's not something coming from an external device, it's just...copy.

Issue #544, opened 2024-02-10
shouse: blastoff support


jotego:

jotego:
@gyurco this was the issue for missing games in shouse core

gyurco:
Here's the PR for the games to start:
https://github.com/jotego/jtcores/pull/587

You don't have to search for the video issues, it'll come as you boot a game. Splatterhouse might be a very good boy to work so good, the others must be the nasty ones.

jotego:
Absolutely, the video problems are all over the place. It is almost a miracle Splatter House looks that good!

I'll close this issue and open specific ones if I find a boot or stability problem in a game.

PR #587, opened 2024-03-23
Add missing protections, make the $a6 hack work with Shadowland (#544)

gyurco:


jotego:
Thank you! As you said, there are many GFX problems. I'll get to them.

gyurco:
I think it also need an 'invert flip' flag, too (see Galaga 88). Probably the first word in the header can be used for that (bit2 - xor flip, bit 0-1 protection type).

jotego:
There must be a hardware setting. If it was game dependent -not in hardware- the MRA would have gotten that information from MAME automatically.

PR #591, opened 2024-03-27
Outrun: use a lfbuf in SDRAM for SiDi128

gyurco:


jotego:
Thank you!

gyurco:
Maybe you can delete the branches created by me, I don't have the rights to do it.

jotego:
I just did. Thanks for asking. I cannot assign better permissions without converting this account to an organization. I'd rather not do that yet.

gyurco:
I don't know if something changed or I was just ignorant, but I was able to delete all my remaining branches. Probably it will work in the future(?)

PR #609, opened 2024-04-11
Sidi128 sta

gyurco:


jotego:
This seems to fix all core but outrun and ngp. toki fails but it has latches and other problems, so I am not concerned about it yet.

I thought that the HDMI clock pin was also connected to a dedicated PLL output. Is the place assignment missing?

I will add waivers for the jitter warnings related to the input clocks in the game PLLs.

Thank you

gyurco:
No, the HDMI_PCLK is not a dedicated output pin. As it's only output and It would require to use another PLL, that's good as it is.
I enabled 96 MHz for Outrun, it's stable, but still glitches here and there (especially in the "stonehenge" track).

PR #613, opened 2024-04-13
Sidi sta

jotego:
I think sidi won't work (even if passing STA) but want to get the files compiled


gyurco:
The PLL location was incorrect. I don't have a real board to test, but I think it should work with the correction.

jotego:
After your change, I was able to consolidate sidi and mist targets again. Thank you very much for spotting it!

PR #617, opened 2024-04-15
modified jotego_20.04.sh script

skutis:
I used the jotego_20.04.sh script after installing ubuntu 20.04 and encountered many errors. I have modified so it passes cleanly on my computer. If anyone else needs it, this is my version of it.

### DIFF
--- a/modules/jtframe/bin/jotego_20.04.sh
+++ b/modules/jtframe/bin/jotego_20.04.sh
@@ -6,63 +6,65 @@
 set -e
 
 # Sublime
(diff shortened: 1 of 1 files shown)


jotego:
isn't it a problem to have so many sudo commands? It forces you to be in front of the computer to re-enter the password, right? I think the sudo command does not ask for a password for a while but eventually it will ask again and the script will pause.

skutis:
> isn't it a problem to have so many sudo commands? It forces you to be in front of the computer to re-enter the password, right? I think the sudo command does not ask for a password for a while but eventually it will ask again and the script will pause.

I did not run the while script at once, but I don’t think there should be a problem.

PR #618, opened 2024-04-15
Fixes #604

### LINKED ISSUE #604: toki: missing MRA generation
Hi @vertrex,

The current `mame2mra.toml` file is a stub. `mame.xml` has been updated to include _Toki_ related machines but the ROM region information is missing from `mame2mra.toml`. This is required to automatically generate MRA files.

Other information that should go in the same file:

- button names
- Optionally shorten DIP switch names or options if those are too long for MiSTer

Thanks

vertrex:
Fixes #604 

Add missing MRA file

### DIFF
--- a/cores/toki/cfg/mame2mra.toml
+++ b/cores/toki/cfg/mame2mra.toml
@@ -1,2 +1,29 @@
+[global]
+mraauthor=["Solal Jacob"]
+author="Solal Jacob"
+webpage="https://ko-fi.com/solalretrogaming"
+twitter="@s_on_vg"
+
 [parse]
 sourcefile=["toki.cpp"]
+skip.Setnames=["tokip"]
+skip.bootlegs=true
+
+[buttons]
+names = [
+    { names="Shoot,Jump" }
+]
+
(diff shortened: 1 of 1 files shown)


jotego:
Thank you. I am not sure why the checks failed to run in the PR, though. That will be my fault in any case.

jotego:
Now that the TOML file is present, the Pocket files should generate correctly too.

jotego:
Pocket and MiST file generation did not work correctly because of the core name. It looks like the support for core names not starting with JT is not complete. I have added it in 65747318
Tested in Pocket: it works but with a garbled line at the top of the screen, probably bad LVBL finish.
Tested in sidi128 too and it works well.

vertrex:
Could it be possible to add support for overwriting the author name, website, twitter (the about xml fields) from  mame2mra.toml fields ? (like it's done with 'jtframe mra --author) so when you generate the release files those information are not lost and author are credited ?

jotego:
Absolutely. I thought that was working. I'll add support for it today. See issue #619

PR #625, opened 2024-04-19
Avoid race condition in verilator test code

skutis:
When I tried jtsim -setname xxx it only worked part of the time on ubuntu 22 and never on ubuntu 20. I debugged it and found that the rm -rf frame.raw in the jtsim script was deleted before the forks creating the frame data files were finished running ImageMagick convert(). I have modified test.cpp to wait until the child processes are completed.


jotego:
Thank you.

PR #638, opened 2024-04-28
SiDi128: high performance effort to fix STA

gyurco:


jotego:
Sorry, I pushed the wrong button. It is merged now.
Thanks

jotego:
It would be good to have this set only for `--nodbg` compilations, in case it makes compilation slow. I'll think about it.

gyurco:
jtshouse compile time is 4:14 with High Performance Effort, 3:28 with Balanced. So yeah, it might worth to turn it off during development.

jotego:
I have implemented it in 4d7fa867
When you compile with `--nodbg` it will use the extra effort, otherwise it will be balanced.

PR #643, opened 2024-05-02
fix missing begin/end in verilog functions for vivado

hansfbaier:
Hi, I am pleased to report that 43 cores already work with MiSTeX on a Xilinx A100T.
The core code has had excellent portability (probably because of the use of the verilator linter).
I only encountered this one minor issue: That vivado wants a begin/end block inside
verilog function declarations.

Kind regards,
Hans

### DIFF
--- a/cores/1942/hdl/jt1942_sound.v
+++ b/cores/1942/hdl/jt1942_sound.v
@@ -157,8 +157,10 @@ jtframe_sysz80 #(.RAM_AW(11)) u_cpu(
 );
 
 function [1:0] bcdir( input cs );
+begin
     bcdir[0] = cs       & ~(hige ? main_wr_n : wr_n); // bdir pin
     bcdir[1] = bcdir[0] &  (hige ? main_a0   :~A[0]); // bc pin
+end
 endfunction
 
 wire bdir0, bc0;
--- a/cores/exed/hdl/jtexed_sound.v
+++ b/cores/exed/hdl/jtexed_sound.v
@@ -127,8 +127,10 @@ jtframe_sysz80 #(.RAM_AW(11)) u_cpu(
 );
 
 function [1:0] bcdir( input cs );
+begin
     bcdir[0] = cs       & ~wr_n; // bdir pin
     bcdir[1] = bcdir[0] & ~A[0]; // bc pin
+end
 endfunction
 
 wire bdir0, bc0;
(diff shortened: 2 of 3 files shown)


hansfbaier:
Current core porting status at MiSTeX:
https://github.com/MiSTeX-devel/MiSTeX-ports/tree/main/cores

jotego:
Thank you for the edits. 
Why don't you make a target for JTFRAME instead and have native support for within the framework?

hansfbaier:
@jotego Thanks for merging! We actually have a mistex target:
https://github.com/MiSTeX-devel/jtcores/tree/master/modules/jtframe/target/mistex
@somhi wrote it. But this will generate (and fail) to build an Altera project.
We use the generated files to build the cores for various FPGA vendors using the python based LiteX build framework.

hansfbaier:
This is how the generated code and the MiSTeX.yaml project definitions look for the jt cores:
https://github.com/MiSTeX-devel/MiSTeX-ports/tree/main/cores/Arcade-jtcores

jotego:
Thank you for the reference links. It looks like somhi's target is a good starting point. Of course, it should not break the other targets. 
Having scripts messing up with the files to compile for a different platform really goes against the spirit of this project. I would like to encourage you or somhi to produce a proper target and make a PR.

hansfbaier:
@jotego I am all for a more elegant solution. However, this might not be very straightforward, since MiSTeX actually targets a variety of different FPGAs and boards:
https://github.com/MiSTeX-devel/MiSTeX-ports/tree/main/mistex_boards
Nevertheless, we currently have a main platform we focus on, which is the Artix 100T with the MiSTeX baseboard.
It would make sense to create a jtframe target for this combination.

jotego:
If there are a lot of them and are mostly of _little interest_ -so to speak- it may make more sense to just add one that uses Vivado so the support for Vivado is directly in the jtcore script.

But I see your point. Somhic did make a PR that had many, many new targets with virtually no users and that would add hundreds of new files to the repository, which is not the idea either. At that time, @gyurco suggested having those targets as submodules, so as not to force other developers to track them and also avoid the file overload on git.

hansfbaier:
Would need to think about situations like this:
20240504_10h41m54s_grim
Use a different file, duplicating a lot of the code, or use `ifdef like here?

hansfbaier:
Hmm and this also fails the verilator linter.

PR #648, opened 2024-05-04
cores/fround/hdl/jt00778x.v: either none or all wires need to be initialized in a combined declaration

hansfbaier:
Another small vivado hiccup.


hansfbaier:
Those CI fails look they are not related to these changes.

jotego:
Thank you

PR #664, opened 2024-05-14
Shouse controls

gyurco:


jotego (approved):

jotego:
Using the header for control selection is the right place indeed.
Thank you!

PR #703, opened 2024-06-14
S18 fixes

gyurco:


modules/jtframe/hdl/ver/test.cpp
```diff
@@ -695,7 +695,7 @@ void JTSim::reset( int v ) {
 }
 
 JTSim::JTSim( UUT& g, int argc, char *argv[]) :
-    game(g), sdram(g), dwn(g), sim_inputs(g), wav("test.wav",48000,false)
+    wav("test.wav",48000,false), sdram(g), sim_inputs(g), dwn(g), game(g)
```
jotego, line 698: Was this order change needed?

gyurco: gcc 14.1 errors out if the member order and the initialization order differs.

Issue #701, opened 2024-06-12
s18: VDP/S16 graphics layer priority


gyurco:
The VDP/S16 rgb output is mixed by an analogue switch.
The mixing is controlled by the [315-5373](https://wiki.pldarchive.co.uk/index.php?title=315-5373) custom chip.
![image](https://github.com/jotego/jtcores/assets/8644936/3e4a31e1-6860-4e9a-9c43-00529a3b5482)

The PLD content translated to Verilog:
```
vdp_sel = vdp_en & vsp_ysn & (!vid16_en
    | (vdp_prio[2] & vdp_prio[1] & vdp_prio[0])  // 7
    | (tilemap[1] & vdp_prio[2] & vdp_prio[1])   // 6,7
    | (tilemap[4] & vdp_prio[2] & vdp_prio[1])   // 6,7
    | (obj_prio[0] & tilemap[1] & vdp_prio[2])   // 4,5,6,7
    | (obj_prio[0] & tilemap[4] & vdp_prio[2])   // 4,5,6,7
    | (!obj_prio[1] & vdp_prio[2] & vdp_prio[1]) // 6,7
    | (!tilemap[0] & vdp_prio[2] & vdp_prio[1])  // 6,7
    | (obj_prio[0] & tilemap[3] & tilemap[1] & vdp_prio[1]) // 2,3,6,7
    | (obj_prio[0] & tilemap[3] & tilemap[4] & vdp_prio[1]) // 2,3,6,7
    | (obj_prio[0] & !obj_prio[1] & vdp_prio[2] & vdp_prio[0]) // 5,7
    | (obj_prio[0] & tilemap[3] & tilemap[2] & tilemap[1]) // all
    | (obj_prio[0] & tilemap[3] & tilemap[2] & tilemap[4]) // all
    | (obj_prio[0] & !tilemap[0]& vdp_prio[2]) // 4,5,6,7
    | (obj_prio[0] & tilemap[3] & !obj_prio[1] & !tilemap[0] & vdp_prio[1]) // 2,3,6,7
    | (obj_prio[0] & tilemap[3] & !tilemap[0] & vdp_prio[1] & vdp_prio[0]) // 3,7
    | (obj_prio[0] & tilemap[3] & tilemap[2] & !obj_prio[1] & !tilemap[0])); // all
```

gyurco:
Is scene simulations are supported in this core?
Write NVRAM writes only a 4k file, it's not the full vram/objram.

[...]

PR #728, opened 2024-07-12
S18 hs

rp-jt:
Changes in interpreting priority of VDP/S16 objects in S18. 
Related to #701


modules/jtframe/hdl/video/jtframe_vtimer.v
jotego, line 1: restore

PR #732, opened 2024-07-16
s18: add wwallyj trackball controls

gyurco:
...and add vertical mouse movement to the dial emulation (second spinner is vertical mouse movement - hopefully it won't interfere with spinner usage in other games).


jotego:
Thank you

Issue #718, opened 2024-06-29
kchamp: music does not stop when fight begins in kchampvs


jotego:
The music stops in MAME. It feels natural it should stop. Setting RECOVERY=0 in the sound CPU did not help.

jotego:
fixed in 207817f by @gyurco. @jtmiki write this one off as done

PR #747, opened 2024-08-01
kchamp: implement sound board reset (#718)

gyurco:


cores/kchamp/hdl/jtkchamp_snd.v
```diff
@@ -61,6 +61,9 @@ reg         ram_cs, pcm_sel, tempo_en;
 reg  [ 1:0] bdir, bc1;
 // sound signals
 wire        nmi_n;
+reg         reset;
+
+always @(posedge clk) reset <= rst | ~snd_rstn;
```
jotego, line 66: The recommendation for reset signals is to use the negative clock edge in order to ease timing. But it does not seem to be critical in practice.

jotego:
Does this fix the music non-stopping bug?

cores/kchamp/hdl/jtkchamp_snd.v
gyurco: Is it something like the clock enables? As my research show it's more like an urban legend than fact...

gyurco:
Yes, it fixes the music non-stopping. Did you experience the opposite?

cores/kchamp/hdl/jtkchamp_snd.v
gyurco: BTW, I can imagine if a signal is used as async reset, then the masking register should be registered at the negative edge, so any change in it cannot cause a glitch in the reset itself (similarly to a gated clock).
E.g.
.async_reset(reset & mask) // mask must be registered at the negative edge, so when reset is high, mask won't glitch, causing erratic reset behavior

However here's the whole signal is registered, so this doesn't apply. Plus it's an 'or' which can cause glitches when the original reset is low.

jotego:
That's perfect. Thanks.

PR #773, opened 2024-08-20
fixed dimming in simson

rp-jt:


jotego:
Thanks for noticing it. This regression comes from a correct inversion applied to jtcolmix_053251, which was not done before. That has removed the need of inversions that were applied externally to 053251 and that did not match the board schematics on punkshot and ssriders.

Instead of taking your extra inversion, which works, I have removed the one I had added at line 66, so it became:

```
assign shd      = |pre_shd;
```

which matches the schematic:

image

Thanks

rp-jt:
Thanks @jotego 
I did that one first but then I noticed that it wasn't working properly for some scenes. For example:

frame_00002

If this options matches the schematics better, then, should we revise 053251 then?

Issue #561, opened 2024-02-24
SF2X/CPS2: launching a core too fast produces visual artifacts


SuperKeeks:
If launching a core very quickly after booting the system, the game shows with visual artifacts. An easy way to repro is to use the bootcore=lastcore option (which launches the game in 10 seconds)

This happens in Street Fighter 2X but it likely affects other CPS2 cores as well.

If waiting 15-20 seconds before launching the core, the issue doesn't happen.

![image](https://github.com/jotego/jtcores/assets/7964724/bc8ee174-9865-4a3a-a275-765cc69dd3c4)

![image](https://github.com/jotego/jtcores/assets/7964724/97a7975d-4247-4e4f-bae4-2e21dd90db26)

![image](https://github.com/jotego/jtcores/assets/7964724/8f5a8f2e-6f90-4efc-a9bb-fa0096a29422)

paulb-nl:
It is not just launching a core too fast. It also happens when loading a certain core before SSF2.

Loading Karnov's revenge in the NeoGeo core will result in massive glitches and flickering in the SSF2 games when loading them afterwards.

Here is a video where SSF2 is working normally and then I start the NeoGeo core:
https://youtu.be/N-a2xj3a_v8

jotego:
The game does not clean up the memories, neither does the hardware. But it should eventually overwrite old values. I wonder if this is related to the game data download being performed through the DDR memory. 
CPS2 is the only jtcore that uses the DDR download in order to speed up the ROM download. I think the NeoGeo core does it too. Maybe it is related to that.
I think you can disable it without recompiling the core by taking away the `address="0x30000000"` text in the mra file.
Could you try that and let me know how it goes, please?

SuperKeeks:
I tried removing the address text. Game loads much slower, but issue still occurs :(

jotego:
> I tried removing the address text. Game loads much slower, but issue still occurs :(

Thank you

paulb-nl:
> The game does not clean up the memories, neither does the hardware

At least the games test Object and Work RAM so when those tests are finished the RAM should have the same contents every time.

jotego:
> At least the games test Object and Work RAM so when those tests are finished the RAM should have the same contents every time.

That's a good point. But if Ryu's silhouette can be seen in those scenes, there must be some wrong data in memory. It could be a MMR (device register), but I cannot think of something that could produce that silhouette effect

SuperKeeks:
[...]

PR #790, opened 2024-09-01
Fix Object RAM clearing for CPS. Issue #561

paulb-nl:
-Add ERASE_AW to clear 0x200000-0x2FFFFF.
-SDRAM write mask was 2'b11 during erasing so it did not clear anything. 
-One small bug left: Last address is not written because erase_bsy goes low too early.


jotego:
Thanks for spotting so many problems. Please give me a few days to study it.

jotego:
Thank you. I have taken the ideas from here to implement it without the new ERASE_AW parameter, which would have affected many files and make the erase feature a bit less obvious. Please see e787541dd28c30798cb6e736a4b4d65e915f6dea

PR #802, opened 2024-09-14
cps1_colmix: adjust layer priority to fix cawing

paulb-nl:
-This fixes the fuel hoses priority bug in cawing. 
-Nemo heads animation is now only visible inside the Nemo letters 
-Lou's gun in 3wonders is now only visible inside square frame


paulb-nl:
These changes are based on the information from CPS RE.

jotego:
I think the change for the Nemo animation is wrong. The animation cannot be read as NEMO if it is only shown inside the letters. Maybe the problem in Nemo is something else.

I am taking the change because the gun frame tilts the balance. Thank you for this fix, Paul.

PR #806, opened 2024-09-16
sync of sprites and shadows in xmen

rp-jt:
This fixes shadow and object not being synced in xmen.
Also, negating the shadow value that goes in the colmix fixes the colour of the images making shd_in. However, this makes shadow_in active low entering jtcolmix_053251 whereas measured in the pcb it appeared to be active high.
Maybe the signal should be negated after entering this module and not before? 

Before changes:
frame_00002

After:


jotego:
I think we measured the Sunset Riders board. Maybe the input signal to the 053251 is negated on the X-Men board. We should check that PCB.

PR #808, opened 2024-09-16
Riders: brightness and shadow

rp-jt:
The shadow information was completely lost in the original version for taking the wrong bit
The brit value was redefined after checking again the schematics

In this version, shadows that weren't drawn by the sprite are visible, for example:

Before:

After:


jotego:
Looks good! But please check whether PunkShot shadows are still visible and report it here. If those shadows did not get broken, I'll merge it.

rp-jt:
@jotego shadows in Punkshot are still visible and seem to be okay after this changes

Issue #765, opened 2024-08-11
xmen: review items


jotego:
Feedback from toya

- [x] Button 3 (special attack) is missing
- [x] OSD option for dip switches/Service mode is missing
- [x] Player shadow is covering enemies
![image](https://github.com/user-attachments/assets/d5e1278b-c729-4792-a1c0-8c017f17f706)
- [x] Magnetos magnet circle is not "circle". See image
![image](https://github.com/user-attachments/assets/8d841d0f-6a9d-442a-9b0c-a87c78b258ea)
- [x] Sentinents mouth has sprite glitch -> maybe related to sprite zero (lime in simpsons)
![image](https://github.com/user-attachments/assets/8013c115-1a39-4a1e-a3b8-4aedc5861bfa)

- [x] Into scene has Visual glitch on top of wolverine
![image](https://github.com/user-attachments/assets/e1d14d7a-1fa2-4794-a63f-d23efa4a9133)
- [x] Explosions have glitches -> same as #741 
![image](https://github.com/user-attachments/assets/fdf81765-ed95-4c4e-8400-13ac1e3485ff)

Toryalai1:
I had a look into what the button names are:
Attack
Jump
Mutant Power

![image](https://github.com/user-attachments/assets/5a0c8aa5-86a8-44b1-8266-cc3a817498fe)

jotego:
thanks

jotego:
when enemies appear from the bottom, a few lines might be missing. Added **scene 11** to check this.

![frame_00002](https://github.com/user-attachments/assets/2ef29266-dd41-4eee-8bb4-2533a3415edc)

rp-jt:
Commit [3127b5b](https://github.com/jotego/jtcores/commit/3127b5b9fb5185192577bbe436fa218d2925252f) fixed remaining issues related to shadow priority

Since all the points mentioned above have been solved, I am going to close this issue

PR #810, opened 2024-09-19
Faster reading of objects in xmen 

### LINKED ISSUE #741: riders/simson: missing lines in objects
Objects may miss lines under some circumstances. It can be seen in the score of _Parodius da!_ and at the end of the second stage of _Sunset Riders_.

Enemies at the top are missing lines
20240727_090653-screen

`CREDIT` characters are missing lines
20240728_062813-screen

rp-jt:
Similar to what was done to fix #741, this fixes horizontal lines for #765

In Xmen, before:

After

Since it's the same chip, I tried applying the same treatment in simson. It gives good results with The Simpsons (before/after):

However, it doesn't work for Vendetta:
frame_00002

As of now, this fix has only been seen in simulation. Still needs to be tested in the game


jotego:
I think this might affect shadows. It may even fix them. Let's check that first.

rp-jt:
Changes added within #835

PR #826, opened 2024-09-30
Independent xmen_colmix

rp-jt:


rp-jt:
@jotego xmen's game.yaml uses:
```
game:
  - from: xmen
    get:
      - "*.v"
```
 So the new file is already included

PR #831, opened 2024-10-03
Fix for sentinel's mouth on xmen

rp-jt:
This change checks if the sprite starts being drawn so much to the left that it actually starts at the right end and, if so, adds a correction in the starting point.

This fixes the sentinel mouth not being correctly aligned(#765 ):

### DIFF
--- a/cores/simson/hdl/jt053246_scan.sv
+++ b/cores/simson/hdl/jt053246_scan.sv
@@ -64,7 +64,7 @@ parameter XMEN = 0;
 reg  [18:0] yz_add;
 reg  [11:0] vzoom;
 reg  [ 9:0] y, y2, x, ydiff, ydiff_b, xadj, yadj, x2;
-reg  [ 8:0] vlatch, ymove, vscl, hscl, hdump_r, hdump_diff;
+reg  [ 8:0] vlatch, ymove, vscl, hscl, hdump_r, hdump_min;
 reg  [ 7:0] scan_obj; // max 256 objects
 reg  [ 3:0] size;
 reg  [ 2:0] hstep, hcode, hsum, vsum;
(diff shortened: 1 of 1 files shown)


cores/simson/hdl/jt053246_scan.sv
```diff
@@ -181,6 +182,8 @@ always @(posedge clk, posedge rst) begin : A
     end else if( cen2 ) begin
         hs_l <= hs;
         dr_start <= 0;
+        hdump_r  <= hdump;
+        if( hdump < hdump_r ) hdump_min <= hdump;
```
jotego, line 186: In general, hdump_min depends on the core. If this was part of JTFRAME, it would make sense to measure it as you are doing. But when you are working on a specific core, you already know the minimum and it is not going to change. So you can just set HDUMP_MIN as a local parameter with the right value.

cores/simson/hdl/jt053246_scan.sv
```diff
@@ -124,6 +124,7 @@ always @* begin : B
     y2     = y + {1'b0,ymove};
     ydiff_b= y2 + { vlatch[8], vlatch } - 10'd8;
     ydiff  = yz_add[6+:10];
+    x2 = x - zmove( hsz, hscl ) - hdump_min;
```
jotego, line 127: As you are only using x2[9], maybe you can use a < or > comparison with HDUMP_MIN and generate a single bit boolean, like left_wrap because the horizontal value is wrapping around the left side of the screen

cores/simson/hdl/jt053246_scan.sv
```diff
@@ -240,7 +243,7 @@ always @(posedge clk, posedge rst) begin : A
                     if( (!dr_start && !dr_busy) || !inzone ) begin
                         {code[4],code[2],code[0]} <= hcode + hsum;
                         if( hstep==0 ) begin
-                            hpos    <= x - zmove( hsz, hscl );
+                            hpos    <= x - zmove( hsz, hscl ) + {6'b0, x2[9], 3'b0};
```
jotego, line 246: based on the single bit boolean, you can add a constant like `left_wrap ? HADJ : 10'd0`

jotego (changes requested):
Good change. Please use a fixed value for hdump_min as indicated in the comment. The other change request (left_wrap) is optional, do it if you find it more readable going forward.

Issue #767, opened 2024-08-12
shouse: flip not working


Toryalai1:
When launching splatterhouse on the analogue pocket and mister (fresh SD card), the game loads upside down. The Rotate Screen option in the OSD doesn't work. When I go into Service mode and toggle the flip option, it does flip, but when exiting, it goes back into upside down.  
Also, when you toggle the "Type" option in the Service menu, it also for some reason toggles the flip

Toryalai1:
Related ticket: https://github.com/jotego/jtcores/issues/691

gyurco:
You have to save the NVRAM content to save the flip option inside the service menu.
I don't know if it's there on Pocket.

Toryalai1:
There is no option to save NVRAM

jotego:
The same problem occurs in MiSTer. There must be something wrong about the DIP switch definition in the MRA file for that game.

Toryalai1:
You are correct, same also in MiSTer

irgun6:
Still not working ?

Exort:
Interestingly during the loading of the service menu boot during Ron test it does show up in the correct orientation, then flips midway through.

I can confirm that flipping in the service menu does resolve the issue but as there's no way to save it it's not useful.

There's no Flip option in the core menu, at least when this game is loaded, so there's no workaround for this issue.

jotego:
Implementing the flip hardware in commit 30b45dfb7c0ec52d97741b9a1a3f40ca9148dd4f had the side effect of revealing that horizontal games on Namco System 1 hardware are displayed through a mirror in the cabinet. The default option is to have the screen flip.

Some of them can be flipped in the service menu (like _splatter_) but in others (like _wldcourt_) the service menu fails because of a RAM error.

Not a comprehensive table:

Setname  | Service Menu   | Can be flipped
---------|----------------|--------------------
faceoff  | how to control it?    | Potentially
mmaze    | 
splatter | Works          | Yes
wldcourt | Works          | Yes

Some horizontal games do not appear flipped:

- pistoldm

Maybe the best solution will be to provide default values for the NVRAM with games preconfigured with the right flip option.

gyurco:
I also think the best would be to have a default NVRAM content then.
[...]

PR #833, opened 2024-10-05
jtframe: don't delay the sel signal in ioctl_dump

gyurco:
...can be still asserted while it shouldn't, see #767


jotego:
thank you

PR #840, opened 2024-10-11
Fix for sprites at the screen bottom (#765) 

rp-jt:


rp-jt:
@jotego all cores using jtframe_draw were tested using this changes and no problems were found

Issue #841, opened 2024-10-11
Simson: Objects missing parts


rp-jt:
The drawing glitches for the object lines are not finished at the end of stage 7 of The Simpsons

![31](https://github.com/user-attachments/assets/a35b3f61-2ef2-4584-9812-229992279d44)
![32](https://github.com/user-attachments/assets/97b181b3-a138-49b0-944c-39dbeea95811)

rp-jt:
Change to 96MHz was reversed for sidi,sidi128,mist,mister in 65c76ddf008d907c5d5742c54ccf5bce33e08472

It only stays for Pocket

rp-jt:
@jotego This issue has reappeared in 4c1b9d1f1877d3a054ddbd051178961b37ab21fb

Check scenes 17 and 31

![Image](https://github.com/user-attachments/assets/6202928d-14f2-42fa-b270-4c52b7f08742)

![Image](https://github.com/user-attachments/assets/400c9864-035e-4403-a1bd-97ee3047cb3e)

PR #845, opened 2024-10-15
Faster reading by reversing order also applied to Vendetta

rp-jt:
Managed to ensure Keep_old==0 in both simson and xmen

Reversing the order had to be made using ~ instead of -, due to priority 0 being used.
When trying to reverse the order using -, the object 0 (which previously was drawn before all other objects and never overwritten) was being overwritten due to still being drawn first, instead of last, for -0 remains unchanged. Applying ~, 0 becomes F, which would come last instead of first, overwritting everything else

SHADOW==0 had to be used in simson because Vendetta uses the shadow pen to write black pixels. With the applied conditions, in the obj buffer, these black pixels could not be drawn over something else

This change slightly helps with #841 but does not fix it completely yet:

### DIFF
--- a/cores/simson/hdl/jtsimson_obj.v
+++ b/cores/simson/hdl/jtsimson_obj.v
@@ -172,9 +172,9 @@ jt053246 #(.XMEN(XMEN))u_scan(    // sprite logic
 );
 
 jtframe_objdraw #(
-    .AW(10),.CW(16),.PW(4+10+2),.LATCH(1),.SWAPH(1),`ifdef XMEN .SHADOW_PEN (4'd15), `endif
-    .ZW(12),.ZI(6),.ZENLARGE(1),.SHADOW(XMEN),.SW(2),
-    .FLIP_OFFSET(9'h12),.KEEP_OLD(0)
+    `ifdef XMEN .SHADOW(1),.SHADOW_PEN (4'd15), `endif
+    .AW(10),.CW(16),.PW(4+10+2),.LATCH(1),.SWAPH(1),
+    .ZW(12),.ZI(6),.ZENLARGE(1),.SW(2),.FLIP_OFFSET(9'h12)
 ) u_draw(
     .rst        ( rst           ),
     .clk        ( clk           ),
--- a/cores/simson/hdl/jtsimson_obj.v
+++ b/cores/simson/hdl/jtsimson_obj.v
@@ -172,7 +172,7 @@ jt053246 #(.XMEN(XMEN))u_scan(    // sprite logic
 );
 
 jtframe_objdraw #(
-    .AW(10),.CW(16),.PW(4+10+2),.LATCH(1),.SWAPH(1),.SHADOW_PEN (4'd15),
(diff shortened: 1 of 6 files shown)


modules/jtframe/hdl/ram/jtframe_obj_buffer.v
```diff
@@ -71,10 +72,10 @@ wire [EW-1:0]       old;
 
 assign new_we = wr_data[ALPHAW-1:0] != ALPHA[ALPHAW-1:0] && we
     && ((old[ALPHAW-1:0]==ALPHA[ALPHAW-1:0]  /*||
-         (old[ALPHAW-1:0]==SHADOW_PEN[ALPHAW-1:0] && SHADOW_PEN!=ALPHA && wr_data[ALPHAW-1:0]!=SHADOW_PEN[ALPHAW-1:0] )*/
+         (old[ALPHAW-1:0]==SHD_PEN[ALPHAW-1:0] && SHD_PEN!=ALPHA && wr_data[ALPHAW-1:0]!=SHD_PEN[ALPHAW-1:0] )*/
          ) || (KEEP_OLD==0 &&
-           ( wr_data[ALPHAW-1:0] != SHADOW_PEN[ALPHAW-1:0] || old[ALPHAW-1:0]==SHADOW_PEN[ALPHAW-1:0]
-            || SHADOW_PEN==ALPHA )));
+           ( wr_data[ALPHAW-1:0] != SHD_PEN[ALPHAW-1:0] || old[ALPHAW-1:0]==SHD_PEN[ALPHAW-1:0]
+            || SHD_PEN==ALPHA )));
```
jotego, line 78: add `&& SHADOW==1` for legibility

modules/jtframe/hdl/ram/jtframe_obj_buffer.v
```diff
@@ -59,6 +59,7 @@ module jtframe_obj_buffer #(parameter
 );
 
 localparam EW = SHADOW==1 ? DW-SW : DW;
+localparam SHD_PEN = SHADOW==1 ? SHADOW_PEN : ALPHA;
```
jotego, line 62: delete SHD_PEN

rp-jt:
@jotego Simpsons and Vendetta were tested with these changes and no problems were reported

cores/simson/hdl/jtsimson_obj.v
```diff
@@ -172,7 +172,7 @@ jt053246 #(.XMEN(XMEN))u_scan(    // sprite logic
 );
 
 jtframe_objdraw #(
-    .AW(10),.CW(16),.PW(4+10+2),.LATCH(1),.SWAPH(1),.SHADOW_PEN (4'd15),
+    .AW(10),.CW(16),.PW(4+10+2),.LATCH(1),.SWAPH(1),`ifdef XMEN .SHADOW_PEN (4'd15), `endif
     .ZW(12),.ZI(6),.ZENLARGE(1),.SHADOW(XMEN),.SW(2),
```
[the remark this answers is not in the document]
rp-jt: Changed in current version. Also, removed `.KEEP_OLD(0)` since it is already the default value

Issue #724, opened 2024-07-06
jtparoda: Parodius graphical glitch when using blue bell explosion super power


RndMnkIII:
In the foreground the sprites of the explosion are correctly displayed while the scaling is taking place. But behind it appears the image of the explosion split into two halves horizontally and inverted vertically.

![image](https://github.com/jotego/jtcores/assets/5108152/b9e08746-2d91-4dbf-9a55-3ed0666e46d8)

jotego:
Another example:
![image](https://github.com/jotego/jtcores/assets/1863036/cc7e3098-2b57-4385-964a-0339218accca)

PR #849, opened 2024-10-16
Fix #846 by increasing the width of the object line buffer

### LINKED ISSUE #846: paroda: Cat-boat reappearing horizontally
In Parodius, at the end of the first stage, if you don't destroy the cat-boat, it will disappear through the left border of the screen.
The same boat reappears right after from the right (picture) before disappearing for good this time. 
Probably similar to #813

rp-jt:
This fixes scene 31 (included in #846) 

It does not help with #724, only with the explosion not appearing from one side of the screen to the other (scene 29):

But the repeating parts of the object are still there so maybe it has to do with the mirroring (scene 28):


jotego:
Note that "bandwidth" refers to a different concept. I renamed the title to "width"

Issue #652, opened 2024-05-06
simson/parodius: wrong sprite position


jotego:
Reported by lmn-sama on Discord

> when the Cat-boat is appearing in the 1st level it appears normally from the lower part of the screen, but some of the sprite is appearing from the top of the screen.

rp-jt:
The "hair"-balloon appears right after having disappeared on top of the screen. Correction factor needs further adjustment
![12](https://github.com/jotego/jtcores/assets/168713859/38920022-c6f0-4c06-8d07-209382eb64a1)

rp-jt:
Fixed in dee657c21e2fceef1eb63a4616f5d8d216dec6ea

Issue #742, opened 2024-07-28
riders: wrong sprite?


jotego:
The bottom half of the pole does not look like the right sprite. Check with MAME. It might be related to the protection chip and thus we may need to wait until Furrtek finds out more about it.

![IMG_6276](https://github.com/user-attachments/assets/e8d5d6bf-29ed-4c77-8960-80aebc67fc00)

There are wrong sprites moving around the saloon girls too. This happens in MAME too and it is known to be related to the protection chip
![IMG_6275](https://github.com/user-attachments/assets/f8b043ce-7aea-469a-95f1-9f032506fe6d)

PR #852, opened 2024-10-18
Removed ywrap and yw0 in k44 and redefined inzone. Fixes #761

### LINKED ISSUE #761: riders: ssriders title logo not displayed correctly for large zoom values
This part of the animation is not visible in the core (42941f5d):

(images from mame)

ssriders use the same object chipset than Parodius, so this might be related to #724

rp-jt:
Fixes #761 

Similar to #840, evaluating y[9] for inzone=0 made bigger sprites rejected from appearing on screen.
I removed that condition and removed the restriction of ywrap, just like how it was done in k46. 
This does not seem to interfere with #652

### DIFF
--- a/cores/riders/hdl/jt053244_scan.sv
+++ b/cores/riders/hdl/jt053244_scan.sv
@@ -59,7 +59,7 @@
 
 reg  [18:0] yz_add;
 reg  [11:0] vzoom;
-reg  [ 9:0] y, y2, x, ydiff, ydiff_b, xadj, yadj, ywrap, yw0;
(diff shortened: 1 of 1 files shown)


real-amano:
I guess that this will not fix #724 either. You guessed that this bug could be related.

rp-jt:
No, @real-amano, this ended up being related to issues with vertical zoom and position. 
If that was the case in #742, I guess the extra parts of the explosion should not appear when the explosion is small and centered, but they appear anyway.

I can confirm that this problem keeps appearing in Parodius after this fix.
We have recently discarded several of the theories we had about this, which also gives us more information on how to fix it

PR #854, opened 2024-10-22
Changed horizontal mirror logic in k44. Fixes #851

### LINKED ISSUE #851: paroda: objects in wrong horizontal position
In the second stage of Parodius, in the tunnel where the big walking woman appears, when she is walking back to the left, the "hair" sprites of her headdress exchange positions and stay at a distance. 
This is until she starts walking to the right and they go back to normal.

frame_00002

rp-jt:
Compared the mirror logic we had for k44 (`ghf ^ pre_hf ^ hmir_eff` ) with the one in Mame. 

Both tables of truth were slightly different, so I added a small change. The pre_hflip is influenced by the horizontal mirroring, but not the other way, and that was causing the problem in #851. This commit fixes #851 

Maybe a similar approach with the vertical flip might help with #724, but I need to check it more thoroughly. I tried changing vflip definition in the same way and the scenes stayed the same

### DIFF
--- a/cores/riders/hdl/jt053244_scan.sv
+++ b/cores/riders/hdl/jt053244_scan.sv
@@ -74,7 +74,7 @@ reg  [ 8:0] zoffset [0:255];
 reg  [ 3:0] pzoffset[0:15 ];
 reg  [ 6:0] pri;
 
-assign hflip     = ghf ^ pre_hf ^ hmir_eff;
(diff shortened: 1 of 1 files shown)


jotego:
It is good. Note that the suffix `_eff` in signals stands for effective, so it is meant to be used instead of the original signal. In this case, it ends up used together with the original. No big deal, though.

PR #855, opened 2024-10-24
Paroda: blue bell explosion fix

### LINKED ISSUE #724: jtparoda: Parodius graphical glitch when using blue bell explosion super power

rp-jt:
This change fixes #724 . Some scenes where the extra pieces of the explosion cannot be seen anymore(before/after):
Scene 26
frame_00002

Scene 28
frame_00002

When the explosion is triggered close to the bottom/top of the screen, something similar can be seen coming from the top of the screen when the explosion is at its biggest, crossing from one border of the screen to the other. However, this effect can also be found in the PCB for this circumstance:

To fix it, corrected vmir_eff evaluation, being more restrictive with the conditions and moved vscl definition to the states machine. Right after registering vzoom (state 2), calculation of the correct vscl took one more clock cycle, if vzoom remained the same, there should be no problem in state 3 for calculating vflip. However, if it did change drastically, like in the case of the explosion, time wasn't enough to include the correct y_diff into vmir_eff, and therefore, vflip would be wrong

### DIFF
--- a/cores/riders/hdl/jt053244_scan.sv
+++ b/cores/riders/hdl/jt053244_scan.sv
@@ -87,7 +87,6 @@ always @(negedge clk) cen2 <= ~cen2;
 always @(posedge clk) begin
     xadj <= xoffset + 10'h66 ;
     yadj <= yoffset + 10'h107;
-    vscl <= rd_pzoffset(vzoom);
     hscl <= rd_pzoffset(hzoom);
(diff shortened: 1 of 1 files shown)


jotego:
Thank you

Issue #867, opened 2024-11-10
wc: support for grid iron


jotego:
The game starts booting but gets halted at the grid screen (9e120b393). Pressing 1P&2P during boot goes through the test screen, which seems ok.

The game also has an LED display, according to MAME:

>Gridiron Fight has a LED display on the control panel, to let each player
>choose the formation without letting the other know.

gyurco:
This is something about interacting with the sound CPU.
The main CPU doesn't get the expected value in the s2m sound latch, and goes to an infinite loop code path.

gyurco:
I have the fix for Gridiron, however World Cup doesn't work even with clear master. Is it expected, or am I have something wrong?

jotego:
World Cup has been running all day long. Regenerate the ROM files as I introduced an 8-byte header earlier today.

gyurco:
Yeah, ROM regeneration worked. I'll post the fix for gridiron soon.

jotego:
Thank you!

jotego:
Thanks, gyurco. @jtmiki this issue is completed.

PR #869, opened 2024-11-10
jtwc: main CPU can reset the sound CPU (#867)

gyurco:
Gridiron Fight expects it


jotego:
I had been looking for the sound reset signal on the board a few times without realising I already had it as _MUTE_!
Thanks

PR #870, opened 2024-11-11
Testbench for jtframe_tilemap

rp-jt:
In this testbench, the values obtained in jtframe_tilemap for vram_addr and pxl are checked to ensure they are okay in case any changes are added to the module eventually.

Also, a warning display is added for when the given parameters are not compatible with vram_addr assignment

### DIFF
--- a/modules/jtframe/hdl/video/jtframe_tilemap.v
+++ b/modules/jtframe/hdl/video/jtframe_tilemap.v
@@ -39,7 +39,7 @@ module jtframe_tilemap #( parameter
     HDUMP_OFFSET = 0,  // adds an offset to hdump
     HJUMP        = 1,  // see jtframe_scroll
     // override VH and HW only for non rectangular tiles
-    VW           = SIZE==8 ? 3 : SIZE==16 ? 4:5,
+    VW           = SIZE==8 ? 3 : SIZE==16 ? 4: 5,
     HW           = VW,
     // localparam, do not modify
     PALW         = PW-BPP,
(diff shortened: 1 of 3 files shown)


modules/jtframe/hdl/video/jtframe_tilemap.v
```diff
@@ -39,7 +39,7 @@ module jtframe_tilemap #( parameter
     HDUMP_OFFSET = 0,  // adds an offset to hdump
     HJUMP        = 1,  // see jtframe_scroll
     // override VH and HW only for non rectangular tiles
-    VW           = SIZE==8 ? 3 : SIZE==16 ? 4:5,
+    VW           = SIZE==8 ? 3 : SIZE==16 ? 4: 5,
```
jotego, line 42: undo the extra space to have a cleaner commit

modules/jtframe/hdl/video/jtframe_tilemap.v
```diff
@@ -91,6 +91,19 @@ initial begin
...
 
+`ifdef SIMULATION
+wire vram_len, vram_big;
+assign vram_len = VA == (MAP_VW-VW + MAP_HW-HW);
+assign vram_big = VA >  (MAP_VW-VW + MAP_HW-HW);
+initial begin
+    if( !vram_len ) begin
+        if( vram_big ) $display("WARNING %m: vram_addr has unassigned bits. Adjust parameters to avoid this");
+        else           $display("WARNING %m: vram_addr has bits assigned by two different sources. Adjust parameters to avoid this");
+        assert(vram_len);
```
jotego, line 102: - vram_len will always be zero if you get to line 102, no need to assert it. Just do `$finish`
- avoid using `assert` in jtframe files because some simulators will not recognize it. Does verilator support it?

modules/jtframe/hdl/video/jtframe_tilemap.v
```diff
@@ -91,6 +91,19 @@ initial begin
     end
 end
 
+`ifdef SIMULATION
+wire vram_len, vram_big;
+assign vram_len = VA == (MAP_VW-VW + MAP_HW-HW);
```
jotego, line 96: shouldn't VA include the code width too?

modules/jtframe/hdl/video/jtframe_tilemap.v
rp-jt: Undone in last commit

modules/jtframe/hdl/video/jtframe_tilemap.v
rp-jt: Changed in last commit
Understood. I think assert is not supported by verilator

modules/jtframe/hdl/video/jtframe_tilemap.v
rp-jt: VA is used in this module to indicate the length of vram_addr, which is assigned this way:

````
assign vram_addr[VA-1-:MAP_VW-VW]=veff[MAP_VW-1:VW];
assign vram_addr[   0+:MAP_HW-HW]=heff[MAP_HW-1:HW];
````

where veff and heff come from vdump and hdump.
The code is not included here but in rom_addr, so there is no need to check CW along with VA

rp-jt:
@jotego yes, it catches the bug solved in #863, which caused the vram_addr to be delayed when flip==1

jotego (approved):

PR #871, opened 2024-11-12
wwfss: screen wider and fixed image in Pocket

### LINKED ISSUE #866: WWF Superstars - Analogue Pocket -- Right side of screen shows graphics
The screen is not centered and the right column shows graphic assets. It does this for every Display Mode. Everything is otherwise playable. Cheers.

rp-jt:
Video macros are duplicated in macros.def for wwfss. Due to this, there is a discrepancy between the dimensions of the image generated by the core (248x240) and the one unintentionally defined in the macros (256x240).

Removing the duplicates and making both dimensions match partially fixes #866, where, in Pocket, a visual effect is produced in the right side of the screen due to this mismatch is reported.

It is also reported that the image is not centered. Comparing with MAME (which uses 256x240) and PCB footage, it is observed that a small portion at the right side of the screen is missing. I adjusted the horizontal blanking and the macros to fix this, setting the video to 256x240


jotego:
256x240 is the right one. It should be the same value as in [DD](https://github.com/jotego/jtcores/blob/master/cores/dd/cfg/macros.def).
I think I cut it by 8 pixels because there was a quirk on the right side of the screen with the tilemap (not the objects!). Check if there is something wrong when you scroll around. Thanks.

rp-jt:
@jotego I relocated the blanking to its original position (moving the screen to the left made that sometimes the first scroll pixel could not be charged correctly)

Apparently, the quirks appearing at the right side of the screen were due to the tilemap character layer. Adding 1 to HDUMP_OFFSET in u_char appears to fix it

PR #877, opened 2024-11-19
Assignment of static signals into registers

rp-jt:


cores/s18/hdl/jts18_main.v
```diff
@@ -175,12 +175,14 @@ wire        dial_cs;
...
 
-wire        mwalk, mwalka, ind_coin, play3;
+reg         mwalk, mwalka, ind_coin, play3;
 
-assign mwalk    = game_id[6];
-assign mwalka   = game_id[7]; // In US version switches are exchanged
-assign ind_coin = ~(dipsw[13]^mwalka) & mwalk;
-assign play3    = ~(dipsw[12]^mwalka) & mwalk;
+always @( posedge clk ) begin
+    mwalk    <= game_id[6];
```
jotego, line 181: Note that for this one (mwalk) there is no associated logic so it would require using an additional flip flop. Still, by registering it again in the module, you are allowing the plane and route tool to have it closer to the point where it is used so it helps timing too.
If you had not registered it, it may have needed to travel from a far point, accumulating delay.
As there are plenty of unused flip flops, registering inputs is a good practice.

Issue #889, opened 2024-12-08
flstory: gfx problems


jotego:
- scene 2, the dragon appears partially on the screen sides, it also shows some horizontal black lines

![frame_00001](https://github.com/user-attachments/assets/045b90cf-c797-4853-9d27-93720ac9c6f0)

- scene 5, part of the bubble covers Ptolemy -the main character- partially

![frame_00001](https://github.com/user-attachments/assets/fd54fd02-83ba-454c-8b9d-c80ea815ed43)

- in a long sim `jtsim -video 800 -w 450 -inputs coin.in`, lines tend to flick, especially around the main character. Trying to capture in a scene (like scene 6) does not show the problem but a long sim shows it. It might be the main problem as the black lines in the dragon

![frame_00787](https://github.com/user-attachments/assets/cce5a177-53a5-4260-989e-938c6ecc0985)
![frame_00788](https://github.com/user-attachments/assets/31796d8f-8762-40d7-b107-4aeb08ad2a36)

jotego:
**Background information**

The video schematics should show how everything works. What I gather from them:

- no double line buffer, sprites to be drawn are decided during the active part of the line and drawn during blanking
- as many as 8 sprites per line are drawn. Sprites over 8 will be drawn every other frame (flicker)
- the video hardware is hard to understand, with a complex multiplexor driving the object RAM. In this RAM, there is the object data table written by the CPU, but the hardware also writes to it the object indexes to draw in the next line
- the same memory also holds column scroll values for the background, but that is not used in FLSTORY

gyurco:
I made a PR, the sprites are OK for me. I simplified it a lot by using a single line buffer.
Flipped mode is not done, but I think it's just need some offsetting in both directions.

jotego:
I think this is completed with the two PR's. Other games in the core have problems, but that would be a different issue.
Thanks @gyurco!
@jtmiki, mark as completed.

PR #891, opened 2024-12-10
flstory: fix sprite vertical positions (#889)

gyurco:
Even in flip mode (actually flip is the non-flip).


cores/flstory/hdl/jtflstory_obj.v
```diff
@@ -31,7 +31,7 @@ module jtflstory_obj(
     input             gvflip,
     input             ghflip,
 
-    input       [8:0] vrender,
+    input       [8:0] vdump,
```
jotego, line 34: once it is a single line buffer, it makes sense that vdump is needed instead of vrender.

PR #914, opened 2025-01-07
Increased pcm gain in aliens

### LINKED ISSUE #910: Crime Fighters - Low drum volume
One of the highlights of this game is its music and for some reason the drums are barely audible compared to what I remember. I've compared it with footage from MAME, Final Burn and even real hardware and the drums sound loud and clear on those.

rp-jt:
Fixes #910

In the master version, in Crime Fighters, percussion cannot be heard clearly in comparison with the rest of the music. In Mame, this is easily differentiated when hearing.

I think it can be heard clearer with this gain value


jotego:
How does this affect the other games in the same core?

PR #927, opened 2025-01-13
Test added for sndchain module

rp-jt:


modules/jtframe/ver/sound/sndchain/gather.f
```diff
@@ -0,0 +1,2 @@
+../../../hdl/sound/jtframe_sndchain.v
```
jotego, line 1: Use `$JTFRAME/hdl` instead of the ../../..
I added support for that so it will be easier to change folder location in the future

rp-jt: Changed in last commit

PR #932, opened 2025-01-14
Updated tilemap test to work with simunit.sh

rp-jt:
I have added additional lines in simunit.sh to have the possibility to add further options. In this case, this is necessary to run SystemVerilog with this test.


jotego:
I think you can add the options directly to the gather.f file. From `man iverilog`:

```
COMMAND FILES
       The command file allows the user to place source file names and certain command line switches into a text file in‐
       stead  of on a long command line. Command files can include C or C++ style comments, as well as # comments, if the
       # starts the line.

       file name
               A simple file name or file path is taken to be the name of a Verilog source file. The path starts with the
               first non-white-space character. Variables are substituted in file names.

       -c cmdfile -f cmdfile
               A -c or -f token prefixes a command file, exactly like it does on the command line. The cmdfile may be  on
               the same line or the next non-comment line.
```

If that works for the system verilog option, it is probably easier to go that way.

rp-jt:
@jotego  I tried including the `-g2012` option in the gather.f, however, it ends up with an error.

According to [documentation](https://steveicarus.github.io/iverilog/usage/command_files.html#command-file-format), it seems not all commands can be passed using a command file.

As it is probably a common requirements for more tests, I have fixed this flag in the `simunit.sh` script.

I have also separated `sim -lxt > sim.log` to another line, since the way it was before could lead to error sometimes, printing "PASS" without even running the simulation

modules/jtframe/ver/video/tilemap/test.v
```diff
@@ -208,7 +207,7 @@ generate
 
 		jtframe_tilemap #(
 			.SIZE(SZ),.VA(VA_T),
-			.CW(CW),.PW(PW),.BPP(BPP),
+			.CW(CW),.PW(PW),.BPP(4),
```
jotego, line 210: shouldn't this be BPP?

Issue #931, opened 2025-01-14
paroda: Parodius Sound/Music Issue?


tunnotron3000:
I've been giving the Parodius core a go as it's one of my favourite games. When booting the core from the MiSTer menu, the game seems to function normally; however, if I reset within the core for any reason, when the game loads again, there's either no sound at all or the sound functions strangely. An example is a random piece of music playing in place of an item collection sound or the music fading out long before it should.

I've uploaded a video here:

https://www.youtube.com/watch?v=zOcfcWhDjs0

I'm using the MiSTer Pi Turbo Pack; everything is updated and I've even re-downloaded the game file via the script to eliminate the change that it was corrupted somehow.

Please let me know if you need any more information - thanks!

jotego:
confirmed on 9fae6e3

gyurco:
It's interesting, the main CPU generates an IRQ for the sound CPU, but how's the sound number is passed to it?

gyurco:
Looks like through the PCM chip.

jotego:
@jtmiki this is fixed.

PR #944, opened 2025-01-21
paroda: sound fixes after reset

gyurco:
- clear the sound RAM
- don't allow the NMI handler execute right after reset by pulling it low, so the CPU has to clear it explicitly to allow the next NMI to happen


gyurco:
Fixes #931

jotego:
Thank you!

jotego:
That was a good use of the NVRAM variation of the Z80 module!

Issue #929, opened 2025-01-14
twin16: ghost sprites in miaj


jotego:
As the main character moves around, he leaves _ghosts_ behind. This can be seen in scene 1, where the character on the ground should not be visible. The one climbing up the stairs is the right one.

![frame_00003](https://github.com/user-attachments/assets/90a94993-a424-473c-894f-306c87b1dcef)

``` BASH
swcore twin16 ver/miaj
jtsim -s 1
```

skutis:
Maybe the sprite table is not emptied correctly?

skutis:
On real hardware it is emptied, it depends on how you have implemented the parsed sprite table scanning.

jotego:
A couple of notes from a conversation with skutis:

- objbufinit starts right at blanking. The 1-line delay was not correct
- the value used to clear the buffer is 8'h00, not MAME's 8'hff

jotego:
Skutis looked into this a bit more. If the CPU tries to access the object RAM while DMA is active, it will get locked because DTACKn does not get asserted for oram access if DMA is busy. 

@rp-jt modify twin16_main:

``` VERILOG
jtframe_68kdtack_cen #(.W(5),.RECOVERY(1)) u_dtack(
// ...
    .bus_cs     ( bus_cs    ),
    .bus_busy   ( bus_busy  ),
    .bus_legit  ( bus_legit )
//...
);
```

- include oram_cs as part of bus_cs
- when `oram_cs&dma_bsy`, bus_busy should also be asserted
- make `bus_legit=oram_cs`

This should fix the boot-up test

jotego:
I added the DTACK-dma_bsy locking as it was done in #946 to both twin16 and fround in df8c5e7c7. But I only implemented the write-lock on twin16 because it breaks the graphics on fround.

skutis:
Is this too time consuming to debug or why not planned? Thanks.

PR #946, opened 2025-01-23
twin16: ghost sprites in miaj

rp-jt:
I have added the different changes commented in #929 

In the current version, boot-up tests pass more often than not, but they still appear to fail sometimes due to the objects.
Also, in Dark Adventure and Vulcan Venture sprites flicker sometimes.

In jt00778x_dma, it was originally proposed to use `oram_we <= nx_cpwa[11];` to clear the table every time, however, this brought some problems with objects in Dark Adventure and Vulcan Venture, since they actively use the dma.
By using `oram_we <= (obj_en | objbufinit) & nx_cpwa[11];`, there is always a phase of clearing the table that can be extended if `dma_on==0`. This also works for eliminating the ghost sprites.

In jttwin16_share, I created the signal `dma_busy` so oram_we is not asserted when the the dma is in use. However, I started passing `objbufinit` instead of the actual `dma_busy` because it seemed to work better in terms of flickering appearing less and more often passing the boot-up tests, but this might be wrong.

There's still some work to go

### DIFF
--- a/cores/fround/hdl/jt00778x.v
+++ b/cores/fround/hdl/jt00778x.v
@@ -60,6 +60,7 @@ module jt00778x#(parameter CW=17,PW=10)(    // sprite logic
     input             lvbl,
     input    [PW-1:0] obj_dx, obj_dy,
     input             gvflip,
+    output reg        objbufinit,
 
     // draw module
     output   [CW-1:0] code,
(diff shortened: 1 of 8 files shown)


jotego:
I think objbufinit is shorter than even the DMA clear phase. We need to use dma_busy.
Look at when the clash happens in simulation and whether DTACKN is being held high correctly while dma_busy is set.
Also check the linter error.

rp-jt:
@jotego It is seen in simulation that DTACKn is held high during dma_busy when oram_cs is active

Also, fixed linter issue and changed it to use dma_busy instead of objbufinit in shared.

jotego:
We cannot merge it if it breaks the boot up.

Issue #989, opened 2025-02-28
jtframe crosshair support: clean up tasks


jotego:
- [x] support for specifying an offset to X, Y coordinates using JTFRAME_LIGHTGUN_XOFFSET/YOFFSET
- [x] crosshair overlayed video should have lhbl/lvbl delayed accordingly
- [x] each crosshair should go invisible indepedently after 5 seconds of no mouse strobe inputs
- [x] make crosshair coordinates visible through jtframe_sys_info, and update debug.md accordingly
- [x] Use XOR instead of OR so the crosshair is visible over backgrounds of the same color
- [x] Make the crosshair 3-pixel wide, instead of just one
- [x] mouse-emulation via keyboard seems to be reversed: upside down/leftside right
- [x] support for screen rotation (required for *bronx* in JTFLSTORY)

jotego:
There was a file missing in d8ef6d8c after merging. I added it on 3f1bdcd81. But when I tested this on MiSTer, the video was completely broken. So I reverted the two commits. You can re-apply them on top of master and debug what got broken on MiSTer.

rp-jt:
#998 removed mouse emulation and added support for vertical games using the mouse

jotego:
solved in 3c2e9af58b2ae008645df0253e5990e96b57af60

PR #992, opened 2025-03-04
Crosshair fix broken video

rp-jt:
This PR reapplies the  commits for using the mouse as lightgun and fixes broken video in Mister 

commit 490f47107cf1b65eaf11b414e688d0904f624488 was reverted because its parent commit was wrong messing some file locations, removing modules and undoing changes previously applied.

Apparently, video breaks in Mister when it has to take lhbl and lvbl from within the crosshair module, whether they have been delayed by the crosshair module or the input is directly assigned as an output. I don`t understand the reason behind this behaviour, but video still seems to fit the screen correctly in mister when ignoring the delayed blankings

Related to #989


modules/jtframe/hdl/jtframe_board.v
```diff
@@ -667,27 +674,34 @@ jtframe_dip #(.XOR_ROT(XOR_ROT)) u_dip(
...
-    .rout       ( cross_r       ),
-    .gout       ( cross_g       ),
-    .bout       ( cross_b       )
-
+    .rst          ( rst           ),
+    .clk          ( clk_sys       ),
+    .pxl_cen      ( pxl_cen       ),
+    .pre_lvbl     ( crdts_lvbl    ),
+    .pre_lhbl     ( crdts_lhbl    ),
+    `ifdef MISTER
```
jotego, line 682: it cannot be different for MiSTer

modules/jtframe/hdl/jtframe_board.v
```diff
@@ -667,27 +674,34 @@ jtframe_dip #(.XOR_ROT(XOR_ROT)) u_dip(
...
+    .lvbl         (               ),
+    .lhbl         (               ), `else
+    .lvbl         ( base_lhbl     ),
+    .lhbl         ( base_lvbl     ), `endif
```
jotego, line 686: vs/hs must be shifted too. All video signals must have the same delay. This is not done in the credits module, but it should have been done. Let's get this one right first.

modules/jtframe/hdl/jtframe_board.v
```diff
@@ -667,27 +674,34 @@ jtframe_dip #(.XOR_ROT(XOR_ROT)) u_dip(
...
-
+    .rst          ( rst           ),
+    .clk          ( clk_sys       ),
+    .pxl_cen      ( pxl_cen       ),
+    .pre_lvbl     ( crdts_lvbl    ),
+    .pre_lhbl     ( crdts_lhbl    ),
+    `ifdef MISTER
+    .lvbl         (               ),
+    .lhbl         (               ), `else
+    .lvbl         ( base_lhbl     ),
```
jotego, line 685: the vertical blank is connected to the _horizontal_ one. That's why it fails. It worked on sidi/mist because of the composite sync signal used at the output, which merges the two of them.

jotego (changes requested):

modules/jtframe/hdl/jtframe_board.v
rp-jt: @jotego solved in latest commit

PR #993, opened 2025-03-05
flstory: Lightgun offset in jtframe

rp-jt:


modules/jtframe/hdl/video/crosshair/jtframe_crosshair_draw.sv
jotego, line 1: You already have the coordinates without offset in the previous module (called x_off and y_off, which is misleading). You could pass those coordinates here so we do not need the offset parameters and we save the adders.

jotego (changes requested):

modules/jtframe/hdl/video/crosshair/jtframe_crosshair_draw.sv
rp-jt: @jotego changed in last commit

jotego:
Please add the description of JTFRAME_LIGHTGUN_X/YOFFSET to macros.md

PR #997, opened 2025-03-07
White borders for lightgun detection

rp-jt:
Modified the Mister scaler to add white borders (needed as reference for Sinden lightguns) and reduce the size of the image when signal is enabled.

Border enabling signal is now associated to `debug_bus[7]` at jtframe_mister level. This should be changed so  the signal can be enabled in the OSD for Lightgun games


modules/jtframe/target/mister/hdl/jtframe_mister.sv
```diff
@@ -167,6 +167,7 @@ module jtframe_mister #(parameter
     output  [ 1:0]  mouse_strobe,
     // Lightguns
     output  [ 8:0]  gun_1p_x, gun_1p_y, gun_2p_x, gun_2p_y,
+    output          gun_border_en,
```
jotego, line 170: this will be generated at jtframe_board, from the status signal. You can use the debug_bus for now, but move it to jtframe_board.

modules/jtframe/target/mister/hdl/sys/ascal.vhd
```diff
@@ -160,6 +167,8 @@ ENTITY ascal IS
 		-- Border colour R G B
 		o_border   : IN unsigned(23 DOWNTO 0) := x"000000";
 
+		-- Enable white borders for SINDEN LIGHTGUN
+		gunen    : IN std_logic;
```
jotego, line 171: keep the full name gun_border_en

modules/jtframe/target/mister/hdl/jtframe_mister.sv
rp-jt: Fixed in last commit

terminator2k2:
Hi @jotego i have used your changes in ascal.vhd and sys_top.v to add a menu osd item for the sinden boarders to enable/disable the boarder in some mister console cores and it works really nice. So to make life easier for sinden users would it be ok to try to push these changes into mister-devel ? (just making sure you would be ok with this first as its your changes) I could then do PR's for the seperate console cores to add these in to the osd menu. Ive also created mgl files for these console cores so they can have separate cfg files. Thanks once again for your work.

jotego:
Thanks a lot for pushing this into the main sys, @terminator2k2. That makes a lot of sense. I see the PR [here](https://github.com/MiSTer-devel/Template_MiSTer/pull/88)

The status bit used for the OSD is the one thing I need to check that is compatible with the ones we are using. Our usage [list is here](https://github.com/jotego/jtcores/blob/master/modules/jtframe/doc/osd.md). We have not chosen a value for the sinden yet, but it should fit in the available ones.

Please @rp-jt have a look at the PR too.

modules/jtframe/target/mister/hdl/jtframe_target_info.sv
```diff
@@ -0,0 +1,135 @@
...
+        endcase
+        2: target_info <= { ioctl_lock, ioctl_cart, ioctl_ram, ioctl_rom, 1'b0, ioctl_wr, dwnld_busy, hps_download };
+        3: target_info <= hps_index[7:0];
+        default: target_info <= debug_bus;
+    endcase
+end
+
+endmodule
+
+module jtframe_mister_status (
```
jotego: make a different file for each module

modules/jtframe/target/mister/hdl/jtframe_target_info.sv
```diff
@@ -0,0 +1,135 @@
...
+`ifdef JTFRAME_VERTICAL
+assign {FB_PAL_CLK, FB_FORCE_BLANK, FB_PAL_ADDR, FB_PAL_DOUT, FB_PAL_WR} = '0;
+`endif
+
+assign uart_rx  = uart_en ? USER_IN[1] : 1'b1;
+assign game_rx  = uart_rx;
+assign joy_in   = USER_IN;
+assign uart_en  = status[38]; // It can be used by the cheat engine or the game
+
+// Mouse
```
jotego: take all the stuff that is not related to status out of this module

jotego (changes requested):

modules/jtframe/target/mister/hdl/jtframe_target_info.sv
rp-jt: Already resolved in 61971cf777166711500c824538207e26694dc66d

modules/jtframe/hdl/lightgun/jtframe_lightgun_scaler.v
```diff
@@ -0,0 +1,61 @@
...
+assign y_next = y_scaled[17:8];
+
+always @(posedge clk) begin
+    strobe <= joyana != joya_l;
+    joya_l <= joyana;
+
+    x_scaled <= x_corr[7:0] * W[8:0];
+    y_scaled <= y_corr[7:0] * H[8:0];
+
+    // if (~x_next[9] & joyana[7])
```
jotego, line 46: delete commented code and make a unit test for this

modules/jtframe/hdl/jtframe_board.v
```diff
@@ -251,6 +253,11 @@ wire [SDRAMW-1:0] bax_addr;
 wire LHBLs;
 
 assign sensty    = status[33:32]; // MiST should drive these pins
+assign gun_border_en = status[9];
+always @(posedge clk_sys) begin
+    hide_gunen <= ~lightgun_en; `ifdef JTFRAME_LIGHTGUN_ON
+    hide_gunen <= 0;            `endif
```
jotego, line 259: do not generate `hide_gunen` in the board module. Make lightgun_en an output port instead and let the mister module handle it instead as hiding option is a MiSTer-specific feature.

modules/jtframe/hdl/lightgun/jtframe_lightgun_position.v
```diff
@@ -0,0 +1,61 @@
...
+    output reg [ 8:0] x,
+    output reg [ 8:0] y,
+    output reg [ 8:0] x_abs,
+    output reg [ 8:0] y_abs,
+    output reg        strobe
+);
+
+parameter XOFFSET=0, YOFFSET=0;
+
+`ifndef JTFRAME_RELEASE
```
jotego, line 36: give priority to the final use case in code. Also, when possible write the checks as positive assertions.
In this case, write `ifdef JTFRAME_RELEASE` instead of `ifndef` and make the release code the first thing we read.

modules/jtframe/hdl/lightgun/jtframe_lightgun_position.v
```diff
@@ -0,0 +1,61 @@
...
+always @(posedge clk) strobe <= m_strobe | (a_strobe & gun_border_en);
```
jotego, line 37: gun_border_en is only used for a debug feature. You are introducing a concept outside of this module abstraction by having it. If you want to keep it in the final PR, consider another way of doing this that keeps sharper responsabilities.

jotego (changes requested):

modules/jtframe/hdl/lightgun/jtframe_lightgun_position.v
rp-jt: We had mentioned adding another option in the OSD menu for enabling/disabling the lightgun crosshair.
I will substitute gun_border_en for this crosshair enabling option, so it can always be accessed either in release or debug mode and the `ifdef clause is not needed anymore

PR #998, opened 2025-03-11
Analog Stick controlling crosshair

rp-jt:
Route connecting Analog stick signal and crosshair gun coordinates.
Sinden lightguns send screen position through analog joystick signal.

With these changes, crosshair can be seen correctly where one is aiming with the lightgun. However, crosshair position using lightgun is shaky, therefore producing failures in shooting tests due to the fast changes. This shaky behaviour is not seen when using an actual joystick

JTFRAME_MOUSE_NOEMU was defined since it interferes with the analog joystick detection

### DIFF
--- a/modules/jtframe/hdl/jtframe_lightgun.v
+++ b/modules/jtframe/hdl/jtframe_lightgun.v
@@ -20,6 +20,8 @@ module jtframe_lightgun (
     input         rst,
     input         clk,
     input         vs,
+    input  [15:0] joyana1,
+    input  [15:0] joyana2,
     input  [15:0] mouse_1p,
     input  [15:0] mouse_2p,
     input  [ 1:0] mouse_strobe,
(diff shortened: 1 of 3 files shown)


jotego:
When using the guns, we should not show the crosshair. Maybe leave it on as a debug option but for releases, if the gun is enabled in the OSD menu crosshairs should go off.

rp-jt:
> When using the guns, we should not show the crosshair. Maybe leave it on as a debug option but for releases, if the gun is enabled in the OSD menu crosshairs should go off.

Okay! That is easy since enabling the crosshair onscreen was already linked only to the mouse strobe before. I will change that back in the next commit

modules/jtframe/hdl/jtframe_lightgun.v
```diff
@@ -39,37 +41,45 @@ parameter WIDTH = 384, HEIGHT = 224,
           YOFFSET= `ifdef JTFRAME_LIGHTGUN_YOFFSET `JTFRAME_LIGHTGUN_YOFFSET `else 0 `endif;
 
 `ifdef JTFRAME_LIGHTGUN
-jtframe_mouse_abspos #(.W(WIDTH),.H(HEIGHT),
+`define JTFRAME_MOUSE_NOEMU
```
jotego, line 44: I am not sure whether defining it here will take effect in time. I think you should check for JTFRAME_LIGHTGUN in the same file where JTFRAME_MOUSE_NOEMU is used.

rp-jt: I changed it in the last commit

rp-jt:
> It looks good. Do you want to merge it now or are you planning more commits?

You can merge it

PR #1041, opened 2025-04-23
Undoc instructions in 6801.yaml

rp-jt:
Fixes Compile All, by including missing instructions

### DIFF
--- a/modules/jt680x/ucode/6301.yaml
+++ b/modules/jt680x/ucode/6301.yaml
@@ -268,8 +268,10 @@ ops:
   - { name: RTI,  op: 0x3b,                                                                     cycles: 10 }
   - { name: RTS,  op: 0x39,                                                                     cycles: 5  }
   - { name: WAI,  op: 0x3e,                                                                     cycles: 9  }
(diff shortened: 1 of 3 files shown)


rp-jt:
Excecuting `jtframe ucode --report jt680x 6801` indicates that the BSR instruction was not cycle accurate, so I have also corrected that in the las commit. 
The obtained output before this change was this:
```
  Op code  | Spec | Core |
-----------|------|------|
BSR  :0x8d |  24  |  23  | <
Warning: 1 instructions are not cycle-accurate

```

jotego:
Can you set the UNDOC instructions in a common file to all instead of duplicating it?

Did you change the cycles for BSR to match the data sheet? What does the data sheet specify?

rp-jt:
> Can you set the UNDOC instructions in a common file to all instead of duplicating it?

I have tried adding both `ops` in the common file, but this is not supported within the use of `include` (according to `jtframe ucode --report`)

> Did you change the cycles for BSR to match the data sheet? What does the data sheet specify?

I changed it because the report indicated that instruction as not cycle-accurate and that value fixed the issue. However, the Data Sheet indicates 6 cycles as stated in the master's version for BSR

Issue #1084, opened 2025-06-14
riders -> lgtnfght: correctly implement dimming


jotego:
Currently (5c567cc92) only the Sunset Riders implementation is in place.

PR #1100, opened 2025-06-25
riders/lgtnfght: Dimming

rp-jt:
In the schematics for Lightning Fighters, the signals equivalent to `dim_cmn[1:0]` are conected to VCC 

This PR solves #1084 

We can see the difference for this change here:

comparativas


rp-jt:
> I will check on MAME 

On MAME there is no dimming of the screen when finishing a stage, everything stays bright

Issue #1119, opened 2025-08-02
riders: golfing greats beta


jotego:
- [ ] DIP switch assignment
- [ ] Tile layer palette
- [ ] No sound
- [ ] DAC
- [ ] PSAC line buffer
- [ ] PSAC correctly displayed
- [ ] DMA connection
- [ ] ROM check
- [ ] Color test (service screen)
- [ ] Scene saving

jotego:
We'll leave the DAC support out for now.

PR #1125, opened 2025-08-05
Scene Saving in Riders for Golfing Greats

rp-jt:
Related to #1119 

Added needed JTFRAME_IOCTL_RD so PSAC and its MMR can be saved
Modified `riders_dump.v` so it can adapt to this. Also had to create `dump_addr` within it so the correct address is sent to the different modules considering the offset created in `ioctl_addr` by the psac memory


jotego:
I have uploaded two scenes to master.
This is how scene 1 renders in simulation:

frame_00002

And this is how it looks on the SiDi128

BB39D610-3CA4-4E3F-B458-8777E3E1BEFA

The simulated version looks better because there is less SDRAM contention

PR #1126, opened 2025-08-06
new docker containers

jotego:


.github/workflows/check-sim.yaml
```diff
@@ -0,0 +1,38 @@
+name: Run Jotego Simulations
```
jotego, line 1: Run Jotego Simulations -> Run core simulations

modules/jtframe/devops/build-dockers.sh
```diff
@@ -1,14 +1,121 @@
...
-docker image build --file linter.df --tag jotego/linter .
-
-# Push the images to dockerhub
-docker login
-docker push jotego/jtcore-base:latest
-docker push jotego/jtcore13:latest
-docker push jotego/jtcore17:latest
-docker push jotego/jtcore20:latest
-docker push jotego/linter:latest
+set -e
```
jotego, line 2: include the -e in the first line `#!/bin/bash -e`

modules/jtframe/devops/build-dockers.sh
```diff
@@ -1,14 +1,121 @@
...
+    FAIL=()
+
+    parse_args $*
+
+    if $PUSH_IMAGES; then
+        prepare_builder $BUILDER_NAME
+        docker login
+    fi
+
+    image="jtcore-base"
```
jotego, line 18: Have one line per image as the original file did, instead of repeating the same code over and over

modules/jtframe/devops/build-dockers.sh
```diff
@@ -1,14 +1,121 @@
...
+
+    image="simulator"
+    path="."
+    if build $image $path $PUSH_IMAGES $PLATFORMS; then
+        SUCCESS+=("$image")
+    else
+        FAIL+=("$image")
+    fi
+
+    echo "✅ Builds completed:"
```
jotego, line 66: Do not use emoticons

modules/jtframe/devops/build-dockers.sh
```diff
@@ -1,14 +1,121 @@
...
+
+	while [[ $# -gt 0 ]]; do
+        case "$1" in
+            --push) PUSH_IMAGES=true ;;
+        esac
+        shift
+    done
+}
+
+main $*
```
jotego, line 121: Use `main "$@"` to cover for some cases in argument contents (like spaces, etc)

modules/jtframe/devops/build-dockers.sh
```diff
@@ -1,14 +1,121 @@
...
+        fi
+    fi
+}
+
+parse_args() {
+    PUSH_IMAGES=false
+
+	while [[ $# -gt 0 ]]; do
+        case "$1" in
+            --push) PUSH_IMAGES=true ;;
```
jotego, line 115: Add a default case to catch misspelled arguments and offer help

modules/jtframe/devops/docker-installer.sh
```diff
@@ -0,0 +1,70 @@
+#!/bin/bash
+
+# This script will remove your existing Docker installation and
+# will do a fresh Docker installation, including official plugins
+# such as composer and buildx (required for building Jotego images).
+# It will also set Docker in rootless mode so you won't need to use
+# sudo, either interactively or within scripts.
+# The script is architecture-independent.
+
+set -e
```
jotego, line 10: move to first line

modules/jtframe/devops/docker-installer.sh
```diff
@@ -0,0 +1,70 @@
...
+
+echo "🚀 Docker installation on Ubuntu for Jotego devops..."
+
+DOCKER_GPG_DIR="/etc/apt/keyrings"
+DOCKER_GPG="$DOCKER_GPG_DIR/docker.gpg"
+ARCH=$(dpkg --print-architecture)
+RELEASE=$(lsb_release -cs)
+
+# 1. Remove any existing Docker installation
```
jotego, line 19: make each one of these sections a function with a descriptive name and remove the comments at the beginning of each one

modules/jtframe/devops/docker-installer.sh
```diff
@@ -0,0 +1,70 @@
...
+echo "🔄 Removing old Docker versions..."
+sudo apt-get remove -y docker docker-engine docker.io containerd runc || true
```
jotego, line 21: Always use long arguments in scripts `--yes` instead of `-y`

modules/jtframe/devops/docker-installer.sh
```diff
@@ -0,0 +1,70 @@
...
+
+echo "🚀 Docker installation on Ubuntu for Jotego devops..."
+
+DOCKER_GPG_DIR="/etc/apt/keyrings"
+DOCKER_GPG="$DOCKER_GPG_DIR/docker.gpg"
+ARCH=$(dpkg --print-architecture)
+RELEASE=$(lsb_release -cs)
+
+# 1. Remove any existing Docker installation
+echo "🔄 Removing old Docker versions..."
```
jotego, line 20: Do not use emoticons. Try to keep terminal output as ASCII

modules/jtframe/devops/docker-installer.sh
```diff
@@ -0,0 +1,70 @@
...
+sudo apt-get purge -y docker docker-engine docker.io containerd runc || true
+sudo rm -rf /var/lib/docker /var/lib/containerd || true
+
+# 2. Install dependencies
+echo "📦 Installing dependencies..."
+sudo apt-get update
+sudo apt-get install -y ca-certificates curl gnupg lsb-release uidmap dbus-user-session
+
+# 3. Add Docker GPG key and official repository
+echo "🔑 Adding Docker GPG key..."
```
jotego, line 31: Do not have `...` at the end of each line. It clutters the screen with no real information. Maybe leave it for cases where you know that it is going to take a long time (several minutes) to run the command.

modules/jtframe/devops/docker-installer.sh
```diff
@@ -0,0 +1,70 @@
...
+PROFILE_FILE="$HOME/.bashrc"
+if [[ $SHELL == *zsh ]]; then
+  PROFILE_FILE="$HOME/.zshrc"
+fi
+
+# Add environment variables only if not already present
+grep -qxF 'export PATH=$HOME/bin:$PATH' "$PROFILE_FILE" || echo 'export PATH=$HOME/bin:$PATH' >> "$PROFILE_FILE"
+grep -qxF 'export DOCKER_HOST=unix:///run/user/$(id -u)/docker.sock' "$PROFILE_FILE" || echo 'export DOCKER_HOST=unix:///run/user/$(id -u)/docker.sock' >> "$PROFILE_FILE"
+
+echo ""
```
jotego, line 66: consider using an in-file document when printing multiple lines of text:
```BASH
cat <<EOF
blah
blah
blah
EOF
```

modules/jtframe/devops/jtcore-base.df
```diff
@@ -12,7 +12,19 @@ RUN apt install -yq rsync
 RUN apt install -yq python python3-pip && pip install pypng
 
 # Go
-RUN GONAME=go1.21.3.linux-amd64.tar.gz; curl -LO https://go.dev/dl/$GONAME && tar -C /usr/local -xzf $GONAME && rm -f $GONAME
+RUN ARCH=$(uname -m) && \
```
jotego, line 15: Try to break this a bit into more lines to make it more readable

modules/jtframe/devops/simulator.df
```diff
@@ -0,0 +1,7 @@
+FROM jotego/linter
+LABEL maintainer=jotego@gmail.com
+
+# For convert command
+RUN apt install -y imagemagick
```
jotego, line 5: `--yes`

.github/workflows/check-sim.yaml
```diff
@@ -0,0 +1,38 @@
...
+
+on:
+  workflow_dispatch:
+  # pull_request:
```
jotego, line 5: delete the comments. You can always copy these lines from other workflows

jotego (commented):
@jorgesg82 look at this

modules/jtframe/devops/jtcore-base.df
jorgesg82: But it is already broken in more lines, what do you mean?

.github/workflows/check-sim.yaml
jorgesg82: This file is still in development; it is not meant to be reviewed yet. I commented them to do some testing.

.github/workflows/check-sim.yaml
jotego: still unchanged. Replace _Jotego_ with _Core_

modules/jtframe/devops/build-dockers.sh
```diff
@@ -1,14 +1,84 @@
...
+        prepare_builder $BUILDER_NAME
+        docker login
+    fi
+
+    build "jtcore-base" "." $PUSH_IMAGES $PLATFORMS
+    build "jtcore13" "/opt/altera" $PUSH_IMAGES $PLATFORMS
+    build "jtcore17" "/opt/intelFPGA_lite" $PUSH_IMAGES $PLATFORMS
+    build "jtcore20" "/opt/intelFPGA_lite" $PUSH_IMAGES $PLATFORMS
+    build "linter" "." $PUSH_IMAGES $PLATFORMS
+    build "simulator" "." $PUSH_IMAGES $PLATFORMS
```
jotego, line 22: Good! If you align the arguments by columns, it will be even easier to read. If `$PUSH_IMAGES` and `$PLATFORMS` are global, you can skip passing them.

modules/jtframe/devops/build-dockers.sh
```diff
@@ -1,14 +1,84 @@
...
+#!/bin/bash -e
+
+BUILDER_NAME="jotego-builder"
+PLATFORMS="linux/amd64,linux/arm64"
+
+main(){
+    SUCCESS=()
+    FAIL=()
+
+    parse_args $*
```
jotego, line 10: This one should also be `parse_args "$@"`

modules/jtframe/devops/build-dockers.sh
```diff
@@ -1,14 +1,84 @@
...
+        echo "Images were pushed to Docker Hub."
+    else
+        echo "Images were loaded locally"
+    fi
+}
+
+prepare_builder() {
+    local name=$1
+    if ! docker buildx inspect "$name" >/dev/null 2>&1; then
+        echo "🔧 Creating buildx builder '$name'..."
```
jotego, line 40: Remove the emoticon

modules/jtframe/devops/build-dockers.sh
```diff
@@ -1,14 +1,84 @@
...
+    fi
+
+    build "jtcore-base" "." $PUSH_IMAGES $PLATFORMS
+    build "jtcore13" "/opt/altera" $PUSH_IMAGES $PLATFORMS
+    build "jtcore17" "/opt/intelFPGA_lite" $PUSH_IMAGES $PLATFORMS
+    build "jtcore20" "/opt/intelFPGA_lite" $PUSH_IMAGES $PLATFORMS
+    build "linter" "." $PUSH_IMAGES $PLATFORMS
+    build "simulator" "." $PUSH_IMAGES $PLATFORMS
+
+    echo "Builds completed:"
```
jotego, line 24: Move all the reporting to a separate function. From here to the current end of _main_

modules/jtframe/devops/docker-installer.sh
```diff
@@ -0,0 +1,85 @@
+#!/bin/bash -e
+
+# This script will remove your existing Docker installation and
+# will do a fresh Docker installation, including official plugins
+# such as composer and buildx (required for building Jotego images).
+# It will also set Docker in rootless mode so you won't need to use
+# sudo, either interactively or within scripts.
+# The script is architecture-independent.
```
jotego, line 8: Warn here about `--host` likely not to work in rootless mode

modules/jtframe/devops/jtcore-base.df
jotego: Lines with different RUN commands breaking when we do something else, like

1. determining and verify ARCH as the first one
2. determining the GO name
3. downloading and installing

Performance-wise is basically the same but it will be a bit easier to read

jotego (commented):
Looking much better, thank you. @jorgesg82 just a couple of small things to change.

jotego:
One warning appears while building the images:

>  1 warning found (use docker --debug to expand):
> - LegacyKeyValueFormat: "ENV key=value" should be used instead of legacy "ENV key value" format (line 13)

modules/jtframe/devops/jtcore-base.df
jorgesg82: That is not possible, as you can't keep environment variables between RUN commands. It means that once I define the ARCH or the OS variable, I only can use them within the same RUN command. There are some ways to break it in more commands, but they will be even harder to read

jotego:
> Yes, this is because in the images `jtcoreXX` the PATH environment variable is being declared in the old way, using `ENV PATH $PATH:...`, instead of `ENV PATH=$PATH:...` If you want to remove the warning I can change it

Yes, let's bring everything up-to-date with the requirements of the current docker version

jotego:
do not use capitals in directory names

jotego:
pull the branch as I fixed a problem in `jtutil mra` related to making the full list of cps2 used zip files.

jotego:
The jobs appear listed as `Simulate (commnd, commando, yiear, yiear)`, can you make it look `
Simulate (commnd/commando, yiear/yiear)` so the core and the setname are linked with a / instead?

PR #1175, opened 2025-09-09
riders/glfgreat: psac compressed tilemap moved to BRAM

rp-jt:
I haven't used `unless: [ POCKET ]` in mem.yaml, since it should not be included  in the smaller FPGAs either and the Pocket's SRAM is not implemented in the core yet

Is it possible to use `unless: [ NOPSAC | POCKET ] ` in `mem.yaml`?

### DIFF
--- a/cores/riders/cfg/mem.yaml
+++ b/cores/riders/cfg/mem.yaml
@@ -112,7 +112,7 @@ bram:
       rw: true
   # PSAC Tilemap 2x2
   - name: t2x2
-    unless: [ NOPSAC ]
+    unless: [ NOPSAC, POCKET ]
     data_width: 16
     addr_width: 18
     sim_file: true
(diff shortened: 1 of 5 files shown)


jotego:
The `unless` takes an array. So the syntax would be: `unless: [ NOPSAC, POCKET ]`

rp-jt:
> The `unless` takes an array. So the syntax would be: `unless: [ NOPSAC, POCKET ]`

Ok. I just added it in the latest commit

> The disadvantage of using mem.yaml is that it forces you to set `DW:16`. In practice, it may not matter because of how BRAM works. Just keep an eye open in case the synthesizer fails to fit the design in MiSTer or SiDi128 after the change.

I already compiled it for SiDi128 and there seemed to be no problem. I'm compiling it for MiSTer just in case

PR #1176, opened 2025-09-09
minor: Regression script print help

rp-jt:
In the master version, if you use `run_regression.sh --help` the script will respond as if you are using too few arguments.

You would need to add more useless arguments to get it to print the help text. Changing the order, `print --help` works as it should


jotego:
make it work with both `--help` and `-h` please

jotego (approved):

Issue #1127, opened 2025-08-07
Wonder Planet graphical issues


wwark:
See the screenshot !

<img width="1152" height="1080" alt="Image" src="https://github.com/user-attachments/assets/84f7e10e-3b99-4473-b828-62834dbd8827" />

Thanks

jotego:
Please @wwark check that you are using the latest combination of .rom .mra and .rbf files

wwark:
I found that is the issue after removing the dip . the behavior is correct.
If I set flip screen you will have this screen.
 
<img width="618" height="602" alt="Image" src="https://github.com/user-attachments/assets/a23e53a7-6dc3-4348-8c3a-a053ed6ab615" />

Could you please reopen the issue ?

jtmiki:
As wwark said if dip flip screen sets ON, there are graphical problems.

<img width="614" height="768" alt="Image" src="https://github.com/user-attachments/assets/e712df91-9024-4ec4-9fda-e1944c6701a1" />

PR #1193, opened 2025-09-17
Better implement flip in jtframe_tilemap

rp-jt:
PR for fixing #1127 

First commit seems to fix scroll and fix position. I still need to test this in more cores that use `jtframe_tilemap`

Additionally, it seems that objects are not drawn in the last 8 pixels of the screen or so when `flip==1`. 
Also, objects look somehow "shaky" in karnov/karnov, but I haven't checked if that is normal

### DIFF
--- a/modules/jtframe/hdl/video/tilemap/jtframe_tilemap.v
+++ b/modules/jtframe/hdl/video/tilemap/jtframe_tilemap.v
@@ -74,7 +74,8 @@ module jtframe_tilemap #( parameter
 );
 
 reg   [DW-1:0]    pxl_data;
-reg [PALW-1:0]    cur_pal, nx_pal;
+reg   [CW-1:0]    fcode;
+reg [PALW-1:0]    cur_pal, nx_pal, flip_pal;
 wire              vflip_g, xhflip;
 reg               hflip_g, nx_hf;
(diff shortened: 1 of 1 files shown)


jtmiki:
If this change affects other cores, please specify which ones.

rp-jt:
> If this change affects other cores, please specify which ones.

Cores that use `jtframe_tilemap.v`: 
- 1942         [Vertical]
- dd/dd2
- flstory      [vertical: bronx/rumba]
- fround     [flip not used]
- gaiden
- karnov     [Vertical: wndrplnt]
- paclan
- rungun   [flip not used]
- thundr
- twin16   [flip not used]
- wc
- wwfss

@jtmiki could you check in today's compile all if any of them break when you set `flip= 1` like it happens in karnov ? (you can skip the ones where flip is not used)

jtmiki:
I have checked them and only Rolling Thunder and Genpei Toumaden of thundr core has a same problem as Karnov core.
They freeze the screen when FLIP is switched on, but these games do not require FLIP in the first place. Other games using the same core show no change when FLIP is activated and operate normally.
20250919_102844-screen
Rolling thunder

20250919_104531-screen
Genpei toumaden

jotego (approved):

PR #1206, opened 2025-09-25
[WIP][Analogizer] Update module to support PSX controllers

sduponch:
Mostly a test to update Analogizer SNAC PSX.

Probably wont work as i can not modify qsf to add analogizer.qip for later module update.


jotego:
JTCORES do not use qip files.

PR #1210, opened 2025-10-01
Jtmiki patch 6 several changes for regression

jtmiki:
I have changed several commandos for regression.


cores/cps1/ver/dynwar/reg.cab
```diff
@@ -1,4 +1,4 @@
 1185
-4 coin
-4
+1 coin
+6
 1 1P
```
rp-jt, line 4: In reg.yaml, I think dynwar is reading other cab file for inputs

PR #1228, opened 2025-10-15
Shadow used in riders_obj. Fixes #1220

### LINKED ISSUE #1220: tmnt2: graphic problem
This screenshot is from a YouTube video.

Image

When playing this game on the MiSTer, at scene 4, if the protagonist or enemies move into the area where the spotlight is shining, they don't appear as shown in this image. Instead, they become completely invisible, as if hidden behind a wall.

Commit: 701515f

rp-jt:
Fixes #1220

### DIFF
--- a/cores/riders/cfg/macros.def
+++ b/cores/riders/cfg/macros.def
@@ -39,6 +39,7 @@ NODIMMING
 
 [mist|sidi]
 NOPSAC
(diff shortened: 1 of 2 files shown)


rp-jt:
@jtmiki tested the different games in riders with this change and everything looks fine!

PR #1248, opened 2025-11-04
ngp/ngpc: power button needs 3 seconds

### LINKED ISSUE #1232: ngp:  prevent turning off by accident
In the ngp and ngpc cores, we can simulate turning off the console by pressing the coin button. This is necessary, for example, if you want to change the game you're playing without going through the console setup menu again.

A user has requested a way to disable this button. Since it is of common use in other games, it is easy to click by accident, especially in Pocket handheld mode, making you lose the progress midgame

Putting this button behind a menu option should not be the solution, since it is fundamental to the console's operation.

Another possibility is adding an option so that "turning off" becomes a button combination, instead of the coin button by itself

rp-jt:
The added module counts the frames while `coin[0]==0` (pressed).
If pressed long enough (around 3 seconds), it will trigger `pwr_button`, to turn on or off the console core. After this happens, for it to work again, the button needs to be unpressed and pressed again for the time said

This long pressing system is more similar to how it worked in the original hardware and helps to avoid clicking it by mistake while playing

Closes #1232

### DIFF
--- a/cores/ngp/hdl/jtngp_game.v
+++ b/cores/ngp/hdl/jtngp_game.v
@@ -27,11 +27,9 @@ wire [ 7:0] snd_latch, main_latch, ioctl_pal, ioctl_main,
 wire [ 1:0] cpu_we, shd_we;
 reg  [ 7:0] st_mux;
 reg  [ 3:0] cart_size;
-wire [ 1:0] pressed_b;
-reg         pressed_l, pwr_press;
 wire        gfx_cs,
             flash0_cs, flash0_rdy, flash0_ok,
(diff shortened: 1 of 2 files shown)


jotego:
I think the way it is written now, with the reference to jtframe_crosshair_disable and the always block at line 78 will make it a bit hard to read in the long run. If we modify the crosshair module in the future, we may also break something here. I think it is better to keep these two things separated.

Please create a module for this that directly instantiates jtframe_countup instead and move the always block inside it too.

jotego (approved):

PR #1258, opened 2025-11-10
Update SDRAM refresh rate calculation constant name

asturur:
I wanted to simply comment on this, but then i thought just open a PR is less noisy than a comment to track and fix.

a 128khz frequency does not match 64us anymore, and so i thought i would rename it.
Then 2 things happened:
- i have to switch to ns because is 7.8125 us
- i also thought that frequency of 78125ns sound bad so i added period in the name

Just close it if you prefer the current name.


jotego:
Thanks!

PR #1310, opened 2025-12-30
Prmrsocr 54539 pin fix

cartman135:


cartman135:
Conflicts resolved. @jotego

PR #1320, opened 2026-01-12
Create reg.cab for glfgreat

jtmiki:
I put an order to start the demo video.


rp-jt:
I think maybe making the video a bit longer would be a good option as well, since the demo should have sound and it shows the different layers (even though it is still broken at the beginning). However, I see in the regression folder that the current video for `Golfing Greats` is shorter than those of the other games of the same core, even though in the `reg.yaml`, we see that `glfgreat` is the only game with `video: 2200`, so it should be the other way.

On the other hand, if you want to start the game completely, this requires a bit more inputs for this game, because you have to choose between several options. If I am not mistaken, you have to press: coin-start-start-B2-B2-B2-B2, and after that, the game actually starts.

Also, don't forget to add the `inputs: reg.cab` line in this core's `reg.yaml`  if you want to go this way

jtmiki:
I have changed the order and I add the inputs: reg.cab in cfg.

Issue #825, opened 2024-09-29
ngp/ngpc: support for saving


jotego:
The game [Evolution](https://en.wikipedia.org/wiki/Evolution:_The_World_of_Sacred_Device) saves the progress. There is at least one user demanding this feature.

Pixelpunker:
Please don't forget this issue. It makes certain games like Sonic Pocket Adventure unsolvable.

jotego:
I assume that saving works and data is stored and restored correctly until you power off the FPGA device. Then the saved data is lost. In order to enable permanent saving, these tasks are needed

- [ ] Find out how per-game data saving is implemented on other cores for MiSTer and Analogue Pocket, both from the user and developer points of view
- [ ] Find out whether the second flash chip can be used to store data. Currently, only the first flash chip supports saving on the core
- [ ] Move the writeable flash page to an **ioctl** block in `mem.yaml`. This implies expanding the flash controller to access both SDRAM and BRAM
- [ ] Implement per-game data restoring, possibly expanding JTFRAME to support it
- [ ] It may be needed to lock writting to this page during the console boot up. There must be a mechanism to unlock it triggered by the games

jtmiki:
I have tested following color games.

**Saved automatically**
 - [x] Biomotor Unitron
 - [x] Dark Arms
 - [x] Densetsu no Ogre Battle Gaiden
 - [x] Dynamite Slugger
 - [x] Fatal Fury
 - [x] The King of Fighters Battle de Paradise
 - [x] Koi Koi Mahjong
 - [x] The Last Blade
 - [x] Magical Drop
 - [x] Memories Off Pure
 - [x] Metal Slug 1st Mission
 - [x] Metal Slug 2nd Mission
 - [x] Mizuki Shigeru Yokai Shashinkan
 - [x] Neo 21
 - [x] Neo Baccarat
 - [x] Neo Cherry Master
 - [x] Neo Mystery Bonus
 - [x] Neo Poke Pro Yakyu
 - [x] Nige-ron-pa
 - [x] Party Mail
 - [x] Picture Puzzle
 - [x] Puyo Pop
 - [x] Puzzle Link 2
 - [x] Samurai Shodown! 2
 - [x] SNK vs Capcom Card Fighter's Clash SNK Ver.
 - [x] SNK vs Capcom Card Fighter's Clash 2
 - [x] Sonic the Hedgehog Pocket Adventure
 
**Uncertain if saved**
 - [ ] Baseball Stars
[...]

PR #1346, opened 2026-02-27
Saves in NGP

rp-jt:


rp-jt:
This PR implements saving for NGP and NGPC as required in #825

PR #1363, opened 2026-03-25
Changed pulse rate for dial emulation sensitivity

rp-jt:
Fixes #1359 

This affects only dial emulation related to keys, by changing the read rate, which was much faster before, leading to not being useful for controlling
Now it sends 1-4 pulses per frame according to sensitivity


rp-jt:
@jtmiki can you test [this compile_all](https://github.com/jotego/jtcores/actions/runs/23848575456) ?

You need to check games that use dial, and see if both dial and dpad emulation work. Thanks!

rp-jt:
With these last changes, I think sensitivity is better for both `arkanoid2` and `calibr50`

Issue #1355, opened 2026-03-15
Caliber 50 sync problems on CRT TV


tdelage26:
Caliber 50 game got problem to sync in RGB 15khz mode (picture appears super dark and muddy on screen). I guess HS backporch value deserves to be tweaked a bit. For the time being in jtcal50_video.v you have HCNT_END( 9'd511 ) , HS_END  ( 9'd480 ) and HB_END  ( 9'd4   ) which gives a backporch value of 511-480+4 -> only 35 clocks which is roughly half what a CRT @15khz expects to properly clamp the signal as far as i know. Maybe lowering down the HS_END to let's say 445 could potentially improve things (that would bring the backporch to 70 clocks)... Just my 2 cents though... ;)

PR #1364, opened 2026-03-25
cal50: sync signals follow schematics

rp-jt:
For #1355 

In calibr50 schematics we find:
image
 But the core was generating 11-8-13 lines. This commit fixes that. Maybe this helps with the sync issues

I have also seen this in the schematics:
image
The core is generating horizontally 36-64-36 instead of 32-64-32. It is also defined `JTFRAME_WIDTH=376` in `macros.def` instead of 384 (as used in MAME and according to schematics), but I haven't changed this one yet, because I assumed it was intentional to reduce black borders. Should I change it back?


skutis:
image
I'm sorry if the intermediate diagrams are confusing, they are there just for information. The important ones are the ones visible at the output of CSYNC.

PR #1380, opened 2026-04-10
Fix compile all

rp-jt:


rp-jt:
Cps3 compilation was failing in mister and sidi128 when trying to get the current commit, before even starting compilation. This was due to it being in another repository. 

Now the commit is obtained in JTROOT instead of inside each core folder. Now, if compilation fails for this cores, it's because of the cores and not because of them being in another repository

jotego (approved):

jotego (approved):

jotego:
Fix the conflicts so we can merge it

PR #1406, opened 2026-05-18
grad3: Fix PCM

### LINKED ISSUE #1403: Gradius III - Sound problems
The issue was already reported on the Discord by another user.

On 2026-05-16 beta release, the voices during weapon choices are incorrectly rendered, with speed inconsistencies & repeating (parasite) noises.

rp-jt:
Included changes suggested in #1403 

Also, corrected the address entering the pcm chip according to schematics, which improved sound a lot:
image

However, it seems voices cannot be heard when starting the stage or getting a power up ( [examples](https://www.youtube.com/watch?v=c-JEBYC4LwY&t=197s) ). Also, I think it might still lack a little bit in the intro music, but I'm not sure what

Before changing the address, a voice effect occurred at the beginning of the first level (although it sounded wrong) and looped.


rp-jt:
The PCM banks use byte-sized data buses, however, the offset needed by the SDRAM bank must be in 16-bit word size. 
The offset was wrong because of that and changing it fixes the PCM problems, fixes #1403

rp-jt:
@jotego this PR is ready to review

Issue #1407, opened 2026-05-19
Grad3 issues


rp-jt:
A list of known issues or planned improvements to the grad3 core
- [x] PCM (#1403 / #1406)
- [x] Change button layout to match Gradius II (#1409)
- [x] Graphic issue with scores (#1408)
- [x] Random death (#1413)
- [x] Credits (#1415 / #1416)

jtmiki:
Reported by @Adel on Patreon

```
I just wanted to let you know Gradius 3 is still having the invisible bullets/enemies.
I managed to get a random death at the very end of a boss battle right in the stage transition. 
Super strange lol I thought I would let you know because not many people can get far enough to test this game. 
I managed to get to the middle of stage 3
```

PR #1409, opened 2026-05-19
Reorder gradius3 buttons

rp-jt:
Reorder buttons to match those set in Gradius II/Vulcan Venture. 

This way, `Missile` and `Shoot` are next to each other in the controller, so they can be easier to press simultaneously


rp-jt:
Contributes to #1407

PR #1439, opened 2026-06-22
feat(): Add double dribble core

asturur:
## Description

This is a ground up implementation of the board labeled GX690 for double dribble.
It is mostly based on assembling known chips from the JT repo using original schematics

Here are the used schematics
cores/ddribble/doc/Double_Dribble_Back.pdf
cores/ddribble/doc/Double_Dribble_Front.pdf

The board has 3 cpu 6809E that talk between them with a pool of shared ram, 2 konami 005885 chips and an ym2203 and a vlm5030, plus the usual colmix and dacs and too many small logic elements around the board.

The core has been written with the help of AI guiding me through schematics, understanding what ls75 and ls138 and ls393 does, with the simulation tools i followed a path that brought me to video and then when things started to be weird i tried to compare mame behavior with what already was done from schematics to see if there were issues.

Most of issues i had was a missing 05885 reference design that was game agnostic, and sound going silent, and then sound going crazy.
All of those were mostly misread inverted clocks, misunderstood labels or wrong clocks.

## How the core plays

The core seems to play just fine.
Controls works, graphic works, there are a bit of stray pixels on a line of sprites that is attached to the crowd, the cheerleaders, but for there rest all seems in place. Since the 05885 works with offsets ( like Jackal and ironHorse do as well ) there could be 1 px of offset that is wrong somewhere, but i couldn't see anything weird.
Audio seems ok, same as the original video i saw.

## Known issues

In theory this board has the same colmix chip of jtcastle, but i couldn't make it work, so i rebuilt it from schematics. It works but if it was a drop-in that would be better

the 05885 needs to run at 48Mhz and i would like to rework it to run at 24Mhz so that all the board is on a single clock and is just simpler

Sound filtering has been eyeballed with AI parsing schematic, the R/C network is complex for my brain and i need some expert to nail down frequence filtering

There are 3 capacistor that are actively turned on/off by the ym2203 and i do not know how to model that, so i didn't.


asturur:
Links to a journal of investigations and fixes for anyone to question the method
cores/ddribble/doc/fixes_journal.md

asturur:
reserved

PR #1445, opened 2026-06-28
Feat(): Arbalest and Metafox

### LINKED ISSUE #1442: feat(): Extend caliber 50 work to metafox and arbalester
Those 2 games, metafox and arbalester run on seta hardware that is slightly different from caliber 50.
It should be possible to run the games with little effort.

asturur:
Arbalest and metafox are 2 very similar games based on a different configuration of the seta hardware, the caliber 50 arcade board.

This core to work required some updates on shared modules that need to be tested once compiled, and the game themselves need to be checked for exact offsets.

I ll update the PR description and turn it non draft as soon as i also tested the affected games ( kiwi/cal50)

gameplay video from original pcb ( audio/video compare):

https://www.youtube.com/watch?v=CRsELrussJU from second 40

videos from the core

https://github.com/user-attachments/assets/260e193b-c240-48d3-b474-583640a20deb

https://github.com/user-attachments/assets/a2ea0f48-3a26-4750-bee1-c08dc51f3664

( the videos ran slow because i encoded them wrongly )

close #1442


asturur:
I turned ON ready for review because i need to grab the compiled caliber50 and jtkiwi

asturur:
i have been told the audio looping on music with samples seems off.

asturur:
@jotego your changes to cal50 and some let go on my side make this PR way thinner it was before. good news. I ll see what else i can do to avoid extra params.

jotego:
I cleaned up it a bit and tested it. I think there are problems with the music.

The distinct filters for left and right channels come from Caliber 50 and may not apply here. Caliber 50 uses that to use lower quality samples for sound effects and better samples for sound instruments. I have the impression that they either didn't do that this time, or they swapped channels.

You can determine this in simulation or live on the FPGA by adding a sampling frequency output that you can inspect. A 4kHz filter means that the sampling frequency is 8kHz or higher. A 8kHz filter requires sampling frequencies of 16kHz or more. By looking at what sampling frequency ranges are being sent to the left/right channels you can determine whether there is a different in use between both channels or not.

Volume is too low too. Setting the overall gain to 1.9 in mem.yaml will help. But it may also be a symptomp of incorrect filtering.

In Metafox, when you finish playing, there are wrong samples played during the continue count down.

asturur:
i ll check sound

PR #1458, opened 2026-07-06
fix(): Double dribble sprite glitches

asturur:
When i tried to rewrite the k05885 with jtframe primitives somehow my AI ended up serializing drawing of tiles and sprites during a line.
That is of course wrong and the original implementation we had before wasn't doing that either.
This change put it back to normal random access, both the tile engine and the sprite engine access the rom data when they need.

This is being currently play tested to see if it solves the last 2 glitches of sprites.


asturur:
ok this remove sprite glitches.
No more lines and no glitching red cheerleader

There is still the sensation that the cheerleaders and the coaches are slitting one pixel when running on side or another but can't pin point to a particular issue.

PR #1463, opened 2026-07-09
feat(): Blockout core

asturur:
I tried to spin up this game becuase it didn't have any custom chip and it is basically a motorola 68000 software rendering on a framebuffer, and as audio board it reuses the entire soundboard of double dragon 2.

The game has no scrolling, no sprites, mostly a static screen with software rendered polygons and a proven audio board.

It was quick to create.

image

image

https://github.com/user-attachments/assets/900af6a9-8beb-4e8d-abd1-77acb68c474f


asturur:
voice audio is too low.

asturur:
audio balanced, voice low pass filter removed.

jotego:
Is this one ready for review?

asturur:
The last open item is if i can adpat the outrun/rungun framebuffer instead of rolling my own framebuffer. For the rest i would say, yes is already reviewed and simplified by me 3 times.

jotego:
The jtframe_lfbuf_* family of modules minimizes memory bandwidth by working on one line at a time. This is fine when the driver of the frame buffer is an object engine, or a tilemap. But if the CPU in Block Out can write at any position of the buffer at anytime, this won't work well. The interface will not match the requirements of the CPU. In that case, you need your own BRAM memory for the buffer, like the `gals` code does.

PR #1464, opened 2026-07-10
fix(): Cabal button names, music speed.

### LINKED ISSUE #1461: Cabal, several audio problems
Tried out Cabal today and these were the audio issues I noted right away:

- music: missing tracks (example: only the low part of the bassline plays for the level bgm)
- music: sped up/glitchy (example: at the end of a level and during boss battles it seems to play at least 2x speed)
- volume: appears to be an inconsistent level between SFX and BGM (however could be perceptual due to missing tracks, etc.)

My set up is MiSTer Pi, running audio out of 3.5mm analog output.

asturur:
- reorder button names
- name the clock of music clock48 so that the framework knows how to handle it
- try to give a bigger rsum to the second adcpm channel that hopefully is the one too loud

closes #1461

### DIFF
--- a/cores/toki/cfg/mem.yaml
+++ b/cores/toki/cfg/mem.yaml
@@ -8,7 +8,7 @@ params:
   - { name: ADPCM1_OFFSET,    value:  "(`ADPCM1_START-`JTFRAME_BA1_START)>>1" }
   - { name: ADPCM2_OFFSET,    value:  "(`ADPCM2_START-`JTFRAME_BA1_START)>>1" }
 clocks:
-  clk48:
+  clk:
     - freq: 3579545
       outputs:
         - cen_fm
@@ -23,8 +23,9 @@ clocks:
 audio:
   channels:
     - { name: fm,   data_width: 16, rsum: 10k }
-    - { name: pcm0, data_width: 14, rsum: 10k }
-    - { name: pcm1, data_width: 14, rsum: 40k }
+    - { name: pcm0, data_width: 12, rsum: 40k }
+    - { name: pcm1, data_width: 12, rsum: 20k }
+    - { name: oki,  data_width: 14, rsum: 10k }
 sdram:
   banks:
     - buses:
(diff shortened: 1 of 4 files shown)


asturur:
It fixed only the buttons names for now.

cores/toki/cfg/mem.yaml
```diff
@@ -8,7 +8,7 @@ params:
   - { name: ADPCM1_OFFSET,    value:  "(`ADPCM1_START-`JTFRAME_BA1_START)>>1" }
   - { name: ADPCM2_OFFSET,    value:  "(`ADPCM2_START-`JTFRAME_BA1_START)>>1" }
 clocks:
-  clk:
+  clk48:
```
jotego: This doesn't change anything. As 96MHz is not used, clk48==clk

asturur: yes zero changes. but i tried since i found a thing with ratio changer, i tried to give it a shot.
I m still looking

cores/toki/hdl/jttoki_sound.v
```diff
@@ -255,6 +256,9 @@ always @(posedge clk) begin
         else if (oki6295_irq_n == 1'b0) //~m68k_sound_cs_4
             irq_rst18 <= 1'b1;
 
+        if (!irq_ack && (cabal ? jt51_irq_n : ym3812_irq_n) == 1'b1)
+            irq_rst10 <= 1'b0;
```
asturur, line 260: @jotego this in theory  unstuck an irq that was always active and firing too fast, and making the musing sound too fast. this fixes the FM for cabal and doesn't do anything bad to toki fm.

cores/toki/hdl/jttoki_sound.v
```diff
@@ -383,7 +387,8 @@ jt51 u_jt51(
 assign cabal_fm_snd = (jt51_l >>> 1) + (jt51_r >>> 1);
 
 assign fm   = cabal ? cabal_fm_snd : opl_snd;
-assign pcm0 = cabal ? {cabal_adpcm0_snd, 2'd0} : oki_snd;
-assign pcm1 = cabal ? {cabal_adpcm1_snd, 2'd0} : 14'sd0;
+assign pcm0 = cabal ? cabal_adpcm0_snd : 12'sd0;
+assign pcm1 = cabal ? cabal_adpcm1_snd : 12'sd0;
+assign oki  = cabal ? 14'sd0 : oki_snd;
```
asturur, line 392: This is a proposal.
Divide oki from pcm0/1 so that we have different rsum for each channel and also we avoid padding with zeroes the cabal pcm, maybe allowing for better mixing too. i m about to test this change here.

asturur:
i see you did it yourself

Issue #1215, opened 2025-10-04
rastan sword sound


Huntsman360:
Sorry for my english, i tested various versions of the core and all have a electric sound in sword attack

In mame and in real arcade the sound is the same, in the fpga core not

jotego:
Might not be related, but the service menu shows an ERROR on the sound row, not letting cycle through the sound codes.

<img width="1024" height="768" alt="Image" src="https://github.com/user-attachments/assets/f020f80c-b190-4f19-828e-d69128a817c2" />

Huntsman360:
maybe, the sound of weapon when you don't hit anything is a very iconic sound is that beacause i remember it and when test in mame is the same. Here is strange like a electric metallic sound.

I don't know if other sounds in the game are wrong, i played some time and don't notice anything strange. But dthe main weapon sound is anoying me

real-amano:
Will fix(rastan): correct PC060 sound NMI control  ( https://github.com/jotego/jtcores/commit/29ea1ddf0ca117b765f6a28984fc4c75ef27720d ) fix this issue? Or is that unrelated?

jotego:
I am working on it.

Some progress notes:

I added a simulation-only hack in jtrastan_snd.v that bypasses the real sound CPU/YM/PCM behavior and feeds the main CPU the exact nibble reply stream captured from MAME for reads at 0x3e0003 during the service-mode sound test. Running `jtsim -d SIM_SND_HACK -d JTFRAME_SIM_VIDEO -dipsw 0000 -video 50 -inputs service.cab` with that hack made the SOUND ERROR message disappear and the service screen report SOUND CODE: 00, RAM OK, and ROM OK, which strongly indicates that the failure is inside the real sound-side implementation or its timing, rather than in the main CPU self-test logic itself.

Huntsman360:
Sorry for being a pain, but have you made any progress with the sound bug? Hopefully we'll finally get the sword to sound the way I remember it instead of that annoying screeching noise

jotego:
I have been focused on CPS3 lately. That new core is almost ready now so I will go back to this soon.

Huntsman360:
Thank you very much, Jotego. I've been following the progress of CPS3, and it's looking very promising. Hopefully you'll manage to figure out the issue with Rastan, because these kinds of small problems can be very difficult to track down and fix.

asturur:
this is the sound from mame
[...]

PR #1473, opened 2026-07-15
fix(): Rastan sound error in service mode

asturur:
This is tied to the issue #1215 investigation.

On the original board schematics the cpu 68000 and z80 are on the same clock chain in ratio 2:1 from a generator of 16Mhz we get 8 and 4 Mhz, on the same signal.

Here they were taken with different dividers chain making the 68000 8800Mhz  and out of phase with the z80 4Mhz.

This clears the sound error in simulation,  i ll make an hardware build to see how it behaves, it may not change the sword sound, that could be one of many things including just audio filtering.

### DIFF
--- a/cores/rastan/hdl/jtrastan_main.v
+++ b/cores/rastan/hdl/jtrastan_main.v
@@ -235,7 +235,7 @@ always @(posedge clk, posedge rst) begin
     end
 end
 
-jtframe_68kdtack_cen #(.W(8)) u_dtack(
+jtframe_68kdtack_cen #(.W(12)) u_dtack(
     .rst        ( rst       ),
     .clk        ( clk       ),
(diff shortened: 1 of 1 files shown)


asturur:
it clears in simulation but not on real hardware

jotego:
This caught a bad typo indeed! Thank you!

The clk frequency in this core is 6.671*8=53.368MHz, not 48MHz. The num/den numbers for the fx68k are derived from a 48MHz master clock, so they are wrong. They were producing 8.9MHz instead of 8MHz because the PLL frequency was not taken into account. Your fix is correct.

We cannot move the fx68k clock generation to mem.yaml but we should move z80's because moving it to mem.yaml takes automatic care of num/den values and of the pll clock frequency.

I will merge this now but if you move z80's cen to mem.yaml, I will take that too!

PR #1476, opened 2026-07-16
refactor(rastan): Move clocks generation on mem.yaml

asturur:
WARNING:
The simulation is still a positive compared to the original failing sound test.


asturur:
I tried to add a gate: [snd] to the clock, just to see if it moves something, but it breaks the sim, i m building anyway to test it on the hardware

asturur:
@jotego ok this is the refactor and not visible regressions

jotego:
Remove the sbaskt file from the PR, please.
I think the Z80 should be on the main `clk` clock domain, not the 24MHz one. That way we can turn on clock throttle for it (by removing the `.RECOVERY(0)` parameter)

asturur:
i tried with and without recovery for the z80 and it didn't really solve anything. but it wasn't on the clk domain, let me try this other setup.

asturur:
how that sbasket file went in? i never touched it.

PR #1479, opened 2026-07-17
refactor(): Rastan pc060 rebuild

asturur:
I was trying to fix the rasta service mode audio test.
At the end i found a pc060 reverse engineered by furrtek with full verilog.
The full verilog didn't run in simulation and i m not sure it could be used as is, so i used AI to correct the one we have with furrtek info.

I also ad to change the clocks again.
From the clock generaged by the dtack, i split it in 2 and i feed the z80 and the pc060, in this way the clocks are identical.

The strong changes to the pc060 and the clock sync make the test pass on the de10

Audio speeded up unexpectedly, this because the 68000 recovery(1) now also send burst clocks to the z80 and the rest. I notice the music seemed faster in certain points. So i disabled the recovery, the game seems to run fine and the audio is more stable, the slowdowns from memory access do not produce any visible change to me.

My current knowledge level does not allow me to say this is better than before, but the test pass.
Ai could have misinterpretd the furrtek HDL or did something else to make me happy, so this change needs a good review.

related to #1215


jotego:
We cannot sync the clocks in that way. The Z80 is going to inherit the wait cycles of the fx68k during memory contention and without recovery for its own contention cycles it will run slower too.

With your other PR, where the fx68k was clocked at 10MHz the sound test problem got fixed already. I would only move the z80 to 48MHz and enable the recovey either at the z80 instantiation or mem.yaml via `gate` for the clock block.

If the service test keeps saying that the communication is fine after that, there is no need for anything else. 

The PC060 rebuild is only needed around PCM sound generation in anycase.

jotego:
The master version as of yesterday 37de44e still has the `SOUND ERROR` message in service mode. So reducing the fx68k speed alone isn't enough. The Z80 must be running more slowly because of the `.RECOVERY(0)`. But you cannot set it to 1 without boosting the clock to 48MHz first.

asturur:
i left incoherent information on discord i ll summarize better here. I m out now

asturur:
Let me try to summarize the whole story.
This does not solve the sword sound first of all, is just a correctness/fidelty fix.

Current master has corrected clock speed and has sound OK only in simulation, in a situation of no ram lag, no recovery needed.

I can't really say what's wrong with the HDL or i understand all the changes, i spent time disassembling the rom, tracing mame and the verilator, to see if i could find the root cause, and this is the least amount of changes that would make the sound check pass. We know i m not good at this stuff, so i looked at all the things i thought could influence the sync, and all permutations of changes to understand if they were all necessary to have the check pass on de10.

I completely understand the point you made on the clock cen synced in this way, but this is the only one working. Gameplay seems identical.

2 things are necessary:
- z80 cen 4mhz generated 1:2 with the m68000 cen at 8mhz, in sync.
- furtek decap info that change our pc060 module
Without one of those, the sound check does not pass on hardware.

** Important note **
furtek decap work is responsibile for 9 out of 10 of the checks to pass.
The clock sync bring it to 10/10

Moving everything on clk48 has been a simplification, and clk48 was chosen so that i could use recovery(0) or recovery(1) on the motorola or on the zilog or both or none. ( on my understanding clk24 is not effective for recovery mode ).

FAILED ATTEMPTS at having the check passing on hardware:
- leave the z80 on the master branch 4mhz setup and move everything to bram ( roms and rams ) to reduce the wait state that would desync the clocks.
- I tried keeping in sync only the m68000 and the pc060 and leave the z80 on a fixed 4mhz clock, nothing
- i tried to have the m68000 with recovery (0) so that would run exact 8Mhz and the z80 recovery 0 at 4mhz, nothing.

It seems that the check is so fragile in sync that they have to be on the same ticking cen, of which one is divided by 2.
In the moment we do that the recovery(1) of the m68000 is going to do the burst also on the z80 derived cen, so the first time i had the check passing in hardware the music was faster,  i thought i botched a clock calculation and was running the z80 too fast but no, it was small burst of recovery(1) triggering

Running with recovery(0) means that the sdram access slow down our 68000 that will have an average clock less than 8mhz, dragging down the z80 clock too at less than 4Mhz, but keeping them in sync. And it also means the z80 recovery time is not accounted at all and is additional.
I could hear faster music with recovery(1) but i can't see slowdowns with recover(0). so i considered this the best branch i have so far.

Both recovery(0) and recovery(1) work fine for the sound check, they both pass.

We may end up trashing this combination, but the take away is:
- there is furtek work to improve the pc060 implementation.
- the clocks on the original board are on the same xtal they are synced by design and they do not need to recover ever
- the sound check is extremely tight in timings and maybe with sdram in place of real roms it just can't pass ever  with recovery clock speeds, and it doesn't have to if it damage gameplay.

I would still look at pc060 changes and try to understand if those are an improvement or just a change for the sake of change.

jotego:
The faster music is not because of the Z80, it is because you passed the wrong clock enable to the FM chips.

asturur:
How much clock needs jt51? 4 and 2 or 2 and 1?

I checked it again it should be 4mhz and cen_p1 2Mhz and it seems to me what i m passing down there.

asturur:
> How much clock needs jt51? 4 and 2 or 2 and 1?
> 
> I checked it again it should be 4mhz and cen_p1 2Mhz and it seems to me what i m passing down there.

... Oh but the FM chip does not need to be in sync, only the z80 has to

asturur:
I marked this 'ready for review'.
I m not touching it anymore i leave to you what to do with those changes.
There are 2 topics:
1)I would like the 060 chip to become a module since is used also on operation wolf and furrtek work make it possibly accurate and fit for a generic reuse.
2) Rastan clocks sync and correctness of this sound check

asturur:
ok this is now updated with latest master, i think it still deserve to be considered as a possibility since it fixes the sound check in rastan but it has to let go of sdram recovery cycles.
I m not sure what is preferred:
- sound check passing, possible slow downs because of sdram access ( to be tested )
- sound check not passing in service mode, cpus able to recover lost time in sdram wait.

At minimum the furtek pc060 decap could be implemented.

This is low priority and shouldn't distract anyone.

jotego:
I took some ideas from here to close this on f8baf8a29.
In the end I accepted taking the cen from the M68000 after I checked that the Z80 cen throttling still worked.

image

Although this may not be robust because of two snd_cen come too close to each other, maybe the Z80 cen_out could be wrong.

For a relatively slow M68000 at 10MHz, I think it will be ok.

asturur:
important is that we had some use of it.

asturur:
Well ok the fix is the PR minus a recovery(0) and the location of the cen divider so i will give myself a pat on the back and be happy about it. 🥳

PR #1480, opened 2026-07-18
feat(): Meikyu Jima

### LINKED ISSUE #1459: feat(): New core derived from vigilantes core to support Meikyu Jima
The hardware is fairly similar it could be booted with a lot of reuse

asturur:
This is the only game that runs on IREM M77 hardware and i derived it from the Vigilante core

Close #1459


cores/vigil/hdl/jtvigil_obj.v
```diff
@@ -111,7 +125,8 @@ always @(posedge clk, posedge rst) begin
                 5: begin
                     { vflip, hflip } <= scan_dout[7:6]^{2{flip}};
                     hsize            <= scan_dout[5:4];
-                    code[11:8]       <= { scan_dout[2], scan_dout[3], scan_dout[1:0] };
+                    code[11:8]       <= CODE_SWAP ?
+                        { scan_dout[2], scan_dout[3], scan_dout[1:0] } : scan_dout[3:0];
```
jotego, line 115: This code swap can probably be avoided in the MRA files by reading the ROM files with a different `sequence`. That is a cleaner way because we should always aim for very simple verilog.

cores/vigil/hdl/jtvigil_obj.v
```diff
@@ -74,6 +79,15 @@ reg         cur_hflip, half_done;
 assign scan_addr = { obj_cnt, sub_cnt };
 assign ydiff     = vf-ypos;
 
+// VSCORE=0 -> no cut. The generate keeps the comparison out of the netlist,
+// as v<0 would be a constant-false unsigned compare.
+generate
```
jotego, line 70: Tell the tool not to generate functions for things that are only used once, and can be summarize as a line. This is a perfect example of that

cores/vigil/hdl/jtvigil_obj.v
```diff
@@ -16,7 +16,11 @@
     Version: 1.0
     Date: 1-5-2022 */
 
-module jtvigil_obj(
+module jtvigil_obj #(
+    parameter [8:0] XBASE     = 9'h180, // sprite X origin
+    parameter [8:0] VSCORE    = 9'd48,  // rows above this get no sprites (0=none)
```
jotego, line 7: Careful with this, it might mean that vertical blanking in this core is above this line?
Or maybe it is an area covered by another layer
So it might be ok as it is, or this VSCORE may not be needed and it should be handled by the colmix to give priority to the other layer.

cores/vigil/hdl/jtvigil_obj.v
```diff
@@ -16,7 +16,11 @@
     Version: 1.0
     Date: 1-5-2022 */
 
-module jtvigil_obj(
+module jtvigil_obj #(
+    parameter [8:0] XBASE     = 9'h180, // sprite X origin
```
jotego, line 6: If the new game uses a different resolution, I guess this makes sense. If not, take it with a grain of salt, it could make more sense to solve this at the vtimer counter generation. I don't know.

jotego (changes requested):

cores/vigil/hdl/jtvigil_obj.v
asturur: it must be noted this was a vigilante value. we just use a different value for the other core.

cores/vigil/hdl/jtvigil_obj.v
asturur: the new game is 384px vs 256 of jtvigil, the main reason why we split the game in its own board, is a different board is just related to the irem m75, is called irem m77 but is just different.

PR #1486, opened 2026-07-21
feat(): tc0030cmd first draft + opwolf wiring

### LINKED ISSUE #1485: feat(): Create a faithful implementation of the taito c-chip
Based on recently relased taito c-chip verilog work, create a shared module that is a fit for jtframe of the chip.

asturur:
This PR vendor ika work on taito c-chip bringing in its current licensing and headers verbatim and adds a wrapper and unit test for the jtframe.

Then it connects the taito c-chip ( tc0030cmd ) to the opwolf prototype enabling the retail version of operation wolf.

close #1485


modules/jttc0030cmd/hdl/IKA87AD.sv
```diff
@@ -0,0 +1,2757 @@
...
+// game "Rainbow Islands". Do NOT wire jttc0030cmd to the rbisland driver
+// until Raki removes this clause. Superman / Operation Wolf / other Taito
+// C-chip games are unaffected.
+//
+// Vendoring lint waivers: the upstream core mixes nibble-width masks with
+// 8/16-bit operands and uses don't-care case patterns by design. Harmless,
+// but noisy under verilator; silenced here without touching the logic.
+/* verilator lint_off WIDTHEXPAND */
+/* verilator lint_off WIDTHTRUNC */
+/* verilator lint_off CASEOVERLAP */
```
asturur: those lines have been added each files.

cores/rastan/hdl/jtrastan_game.v
```diff
@@ -198,4 +207,77 @@ jtrastan_video u_video(
...
+
+jtframe_ioctl_range #(.AW(13), .OFFSET(`JTFRAME_PROM_START+22'h1000)) u_ccepr_dl(
+    .clk    ( clk           ),
+    .addr   ( prog_addr     ),
+    .addr_rel( cc_epr_waddr ),
+    .en     ( prom_we       ),
+    .inrange( cc_eprom_we   ),
+    .din    ( prog_data     ),
+    .dout   ( cc_epr_dd     )
+);
```
asturur: @jotego this part here load the taito cchip rom ( common and per game ) inside the chip private memory. This is what claude said i should have done, i have never seen this so please check.

cores/rastan/hdl/jtrastan_game.v
```diff
@@ -198,4 +207,77 @@ jtrastan_video u_video(
...
+
+always @(posedge clk) begin
+    LVBLl_cc   <= LVBL;
+    cchip_int1 <= LVBLl_cc && !LVBL;
+end
```
asturur: This is a pulse generation. if we pass just ~LVBL inside the chip it doesn't work. I tried to make this logic part of the wrapper, but again claude said no, this is perfect. i think is wrong because likely we need to do it for all games that use c-chip at this point should go in the module wrapper.

cores/rastan/hdl/jtrastan_main.v
```diff
@@ -153,19 +169,46 @@ assign VPAn     = !(!ASn && FC==7 && A[3:1]==5 && RnW);
...
+                boot_addr <= 18'h1ffd7;   // word of byte 0x3ffae (Y calibration)
+                boot_st   <= 1'b1;
+            end
+            1'b1: if( boot_ok ) begin      // romy = ROM low byte @ 0x3ffae
+                gun_yoffs <= 9'h1f8 - {1'b0, rom_data[7:0]};
+                boot_busy <= 1'b0;         // release the CPU
+            end
+        endcase
+    end
+end
```
asturur: This whole block keep the cpu stalled till can read the rom, it needs 2 values, then keep the values.
Since this is not how hardware work, in theory those values are accessed by the lightgun hardware right away from the rom, i would be happy to not have this, and just read from the rom those values and hardcode them somewhere for each game version. like in the MRA or in the code.

asturur:
@jotego i think this is ready for a review.

cores/rastan/hdl/jtrastan_main.v
asturur: i killed it. put constants in headers

cores/rastan/hdl/jtrastan_game.v
asturur: i killed and moved it in mem.yaml + interface changes

cores/rastan/hdl/jtrastan_game.v
asturur: put it in the chip.

asturur:
oh i need to do the debug list

asturur:
IMG_9187

asturur:
something off with the screen, i need to check if prototype does same.

modules/jttc0030cmd/hdl/jttc0030cmd.v
```diff
@@ -0,0 +1,320 @@
...
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License
+    along with JTCORES.  If not, see <http://www.gnu.org/licenses/>.
+
+    Author: Andrea Bogazzi <andreabogazzi79@gmail.com>
+    Version: 1.0
+    Date: 21-7-2026 */
+
+// Taito TC0030CMD "C-Chip" — jtframe wrapper around the IKA87AD core.
```
jotego: Delete 99% of comments from this file. Whatever is worth keeping, let's move it to the module README.md file

cores/rastan/cfg/mame2mra.toml
```diff
@@ -39,9 +52,12 @@ regions = [
     { name="audiocpu", start="JTFRAME_BA1_START" },
     { name="adpcm", start="PCM_START"    },
     { name="pc080sn", start="JTFRAME_BA2_START", width=16, no_offset=true  },
-    { name="pc090oj", start="JTFRAME_BA3_START", width=16, no_offset=true  }
+    { name="pc090oj", start="JTFRAME_BA3_START", width=16, no_offset=true  },
+    { name="cchip_bios", setnames=["opwolf","opwolfa","opwolfj","opwolfjsc","opwolfu"], start="JTFRAME_PROM_START", rom_len=0x1000, files=[ { name="cchip_upd78c11.bin", crc="43021521", size=0x1000 } ] },
```
jotego, line 56: we shouldn't need the explicit `files` part here. Just the region name should suffice.
I am not sure but I think we do not need to list the setnames either. If the other games don't have the region it will work anyway.

cores/rastan/cfg/mame2mra.toml
```diff
@@ -39,9 +52,12 @@ regions = [
...
+    { name="cchip:cchip_eprom", rename="cchip_eprom", setnames=["opwolf","opwolfa","opwolfj","opwolfjsc","opwolfu"], rom_len=0x2000 }
```
jotego: Why the `:` in the name? And why it doesn't appear with the colon in the `order` element at the bottom?

cores/rastan/cfg/mem.yaml
```diff
@@ -7,6 +7,14 @@ clocks:
     - freq: 384 kHz
       outputs:
         - pcm_cen
+  clk48:                    # on the main clk domain (where the C-chip host bus lives)
+    - freq: 12 MHz          # Operation Wolf C-chip MCU (uPD78C11) clock enable
+      outputs:
+        - cchip_cen
+bram:
+  # Operation Wolf C-chip ROMs, loaded from the PROM download region into BRAM.
```
jotego: remove comment

cores/rastan/cfg/mame2mra.toml
```diff
@@ -1,28 +1,41 @@
...
 [buttons]
 names=[
     { machine="rastan",  names="Slash,Jump" },
     { setname="opwolfp", names="Bullet,Grenade" },
+    { setnames=["opwolf","opwolfa","opwolfj","opwolfjsc","opwolfu"], names="Bullet,Grenade" },
 ]
 
 [header]
-info="Byte 0 bit 0: Operation Wolf hardware"
+info="Byte 0 bit0: Operation Wolf hardware; bit1: C-chip present. Byte 1/2: light-gun X/Y offset (signed)"
```
jotego: There's a different syntax to add the bits that will also generate a `jtrastan_header` file to instantiate directly in `jtrastan_game`. That method is preferred

cores/rastan/hdl/jtrastan_game.v
```diff
@@ -39,15 +49,27 @@ assign xram_cs  = ram_cs | vram_cs;
 assign ram_dsn  = main_dsn;
 assign main2snd = opwolf ? main_dout[11:8] : main_dout[3:0];
 
+// Header: byte0 bit0=Op Wolf hardware, bit1=C-chip present; byte1/byte2 =
+// the gun-calibration ROM bytes (rom[0x3ffb0]/rom[0x3ffae]) written in by mame2mra.
 always @(posedge clk) begin
-    if (header && prog_we && prog_addr[3:0]==0) opwolf <= prog_data[0];
+    if (header && prog_we) case (prog_addr[3:0])
```
jotego: Replace with the auto-generated `jtrastan_header`

cores/rastan/hdl/jtrastan_game.v
```diff
@@ -198,4 +220,46 @@ jtrastan_video u_video(
     .debug_view ( debug_view)
 );
 
+// cchip_cen (12 MHz MCU clock enable) and the two C-chip ROM BRAMs
```
jotego: remove all the comments in this section

cores/rastan/hdl/jtrastan_main.v
```diff
@@ -153,9 +161,10 @@ assign VPAn     = !(!ASn && FC==7 && A[3:1]==5 && RnW);
 assign bus_cs   = rom_cs | vram_cs | ram_cs;
 assign bus_busy = (rom_cs & ~rom_ok) | ( (vram_cs | ram_cs) & ~ram_ok);
 assign bus_legit= vram_cs & ~sdakn;
-assign opwolf_gun_x = gun_x + 9'd26;
-assign opwolf_gun_y = gun_y - 9'd6;
-
+// Light-gun offsets come from the header (gun_xoffs/gun_yoffs inputs), derived
+// per set at MRA build time from the same ROM bytes MAME's init_opwolf reads.
+assign opwolf_gun_x = gun_x + gun_xoffs;
```
jotego: register these signals

modules/jttc0030cmd/cfg/files.yaml
```diff
@@ -0,0 +1,11 @@
+jttc0030cmd:
+  - get:
+    # IKA87AD core (vendored, Raki, BSD-2 + clause 3 — see LICENSE.IKA87AD).
```
jotego, line 3: This "vendored" terminology doesn't really apply here.

modules/jttc0030cmd/hdl/IKA87AD_mnemonics.sv
```diff
@@ -0,0 +1,309 @@
+// IKA87AD — NEC uCOM-87AD (uPD78C11) microcontroller core.
```
jotego: I would leave these files as they were in the original without adding extra comments

jotego (changes requested):
Quite good! Just a few changes, especially the header one would be nice to change.

asturur:
you have no idea how many comments i already deleted and how many times i said 'do not add comments' to the tool. is frustrating.

asturur:
image
this needs to be fixed but it seems to me it has the same issue in current master.

cores/rastan/cfg/mame2mra.toml
asturur: this was explained to me as 'because this is the rom name for mame source code and that is how we need to reference it for the mra generation to work'. it doesn't appear with the colon later because we are using 'rename' here.

asturur:
ok timing closure is fixed
alignment is fixed
I m going to do a bunch of play testing of operation wolf and rastan, then try to remove that extra info from the mame2mra file.

asturur:
well in this branch opwolf is now fully broken

asturur:
@jotego beware of the 3 hidden changes:
- timing reports in case of failure ( new scripts )
- header generation changes ( extra bits )
- the SDC file cores/rastan/syn/timing.sdc i added for the c-chip

jotego (approved):

jotego:
what triggers the timing reports? Is it automatically executed by `jtcore` It already printed a slack number that was enough.

asturur:
> what triggers the timing reports? Is it automatically executed by `jtcore` It already printed a slack number that was enough.

In case you get a negative slack and the build fails, the script generate the report and uplaods it so you can read it.
```
TIMING_RPT="$JTROOT/log/$TARGET/$corename.timing.rpt"
            if quartus_sta -t $JTFRAME/bin/jtreport_timing.tcl $corename > "$TIMING_RPT" 2>&1; then
                echo "Detailed timing report written to $TIMING_RPT"
            else
                echo "WARNING: could not generate detailed timing report (see $TIMING_RPT)"
            fi
```

Honestly is just to help me that i can't run quartus on my machine yet.

PR #1489, opened 2026-07-21
kiwi: fix dial reset

### LINKED ISSUE #1488: kiwi/arkanoid2: Coin overflow triggers too easily
If you introduce more than one coin and press start, the `COIN OVERFLOW` screen will appear

Mentioned in #1227

rp-jt:
For #1488


rp-jt:
@jtmiki could you test `arkanoid2` in the Compile All for this PR? Could you check if the sound is right for the games in this core (`kiwi`)?

jtmiki:
I have checked the sounds of all games in this core and they were fine.

skutis:
I should somehow get the rom to the mcu for arkanoid 2. There are mechanisms to extract rom for some MCU:s. I might attempt something like that.

PR #1492, opened 2026-07-22
fix(rastan): Video align for operation wolf and rastand and timing fix 

asturur:
This PR aims at fixing 2 issues with the current rastan/opwolf core.
1) timings on pocket broke in master with the addition of opwolf
2) opwolf makes clear that the video is shifted. If compared with mame both rastan and opwolf are shifted 5px on the left and 1 px down compared to the emulator.

While rastan is not noticeable the difference, opwolf is clearly broken.

Before

opwolf_stack

rastan_stack

After

opwolfp_2804_hostage_stack
opwolfp_4200_map_stack
rastan_before_after

The solution for the timing seems to be to break down the audio with a reg, producing a 1 clock shift in audio that shouldn't be an issue.

### DIFF
--- a/cores/rastan/hdl/jtrastan_obj.v
+++ b/cores/rastan/hdl/jtrastan_obj.v
@@ -139,7 +139,7 @@ always @(posedge clk, posedge rst) begin
             rom_addr <= { code[12:0], ydiff[3:0]^{4{~attr[15]}}, attr[14] };
             half     <= 0;
             dr_busy  <= 1;
-            buf_pos  <= xpos[8:0] + 9'd14;
+            buf_pos  <= xpos[8:0] + 9'd13;
             cur_pal  <= attr[3:0];
             cur_hflip<= attr[14];
             buf_we   <= 0;
--- a/cores/rastan/hdl/jtrastan_scr.v
+++ b/cores/rastan/hdl/jtrastan_scr.v
@@ -70,19 +70,16 @@ module jtrastan_scr(
     output   [ 7:0] debug_view
 );
 
-wire [ 8:0] vdump;
+wire [ 8:0] vdump, vrender_raw;
 reg  [15:0] scr0_hpos, scr1_hpos, scr0_vpos, scr1_vpos;
+
+localparam [8:0] VADJ = 9'd1;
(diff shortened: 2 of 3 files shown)


cores/rastan/hdl/jtrastan_obj.v
```diff
@@ -139,7 +139,7 @@ always @(posedge clk, posedge rst) begin
             rom_addr <= { code[12:0], ydiff[3:0]^{4{~attr[15]}}, attr[14] };
             half     <= 0;
             dr_busy  <= 1;
-            buf_pos  <= xpos[8:0] + 9'd14;
+            buf_pos  <= xpos[8:0] + 9'd13;
```
asturur, line 142: left shift for sprites, a different 1 px bug compared to mame optional compared to the mess of the 5px generic shfif that breaks operation wolf.

cores/rastan/hdl/jtrastan_snd.v
```diff
@@ -78,10 +79,14 @@ assign pcm0_scaled = pcm0_raw * $signed({1'b0,pcm0_vol});
 assign pcm1_scaled = pcm1_raw * $signed({1'b0,pcm1_vol});
 assign pcm0_atten = pcm0_scaled[19:8];
 assign pcm1_atten = pcm1_scaled[19:8];
-assign pcm0 = opwolf ? pcm0_atten : pcm0_raw;
-assign pcm1 = pcm1_atten;
+assign pcm0 = pcm0_q;
+assign pcm1 = pcm1_q;
```
jotego, line 84: see c5f5c4db2

cores/rastan/hdl/jtrastan_scr.v
```diff
@@ -70,19 +70,16 @@ module jtrastan_scr(
     output   [ 7:0] debug_view
 );
 
-wire [ 8:0] vdump;
+wire [ 8:0] vdump, vrender_raw;
```
jotego, line 73: See 1a3c1f304
I think this could have been done by shifting LVBL instead

asturur: shifting LVBL resulted in a full black sim for both games to me. ( you mean changing the timing generation correct? )

PR #1495, opened 2026-07-25
Adds Thunder Cross II (thndrx2) to the JTTMNT core

niknakniknak:
Adds Thunder Cross II (Konami GX073, 1991 — World / Asia / Japan) as a new
`game_id` in the existing `cores/tmnt` core.

Hardware.

- 68000 main, Z80 sound, YM2151, K053260
- K052109 tiles, K051960/K051937 sprites
- K053251 priority, K054000 collision, ER5911 EEPROM, no DIP switches

Notes.

- K054000 is from simson.
- All new logic is gated on `game_id` — TMNT / MIA / Punk Shot code paths are
  byte-identical to before.
- SDRAM layout reworked in `macros.def`.thndrx2's Z80 is 64 KB (old slot
  32 KB) and its K053260 samples are 512 KB (old PCM slot 128 KB); downstream slot
  sizes are preserved so the other three games still fit. Because the offsets
  moved all four games' MRAs need to be regenerated.
- Per-region EEPROM defaults (thndrx2 / thndrx2a / thndrx2j), keyed on `setname`.
- Service Mode DIP (the board has no real DIPs), an FM/PCM balance pass, and an optional FM-in-mono-mode
  centring DIP. All gated to thndrx2.

Testing.

- All three regions boot; graphics correct; played past stage 6 to test K054000, matches PCB
- EEPROM save/restore, service mode, and audio verified.
- Sound balance set from real-PCB recordings (K053260 PCM sits above the YM2151).
- Service-switch behaviour cross-checked against PCB and mame.

Co authored with some help and guidance from @claude


jotego:
Quite nice. Thank you. The code looks decent. I would only change a couple of minor things. Let me push directly the changes rather than ask you for them.

jotego:
I think you need to enable something so I can push on your PR. I don't know if you can do it on an existing PR or create a new one. Can you check it please?

asturur:
It is a setting in the fork 'allow owner to push commits' or something like that.

niknakniknak:
image

this is already on. also added @jotego on my repo. not sure what else it could be, Or a conflicting outstanding merge?

jotego:
I squashed it by mistake. After that I forced another push undoing the squash.

jotego:
Thanks a lot. This is a good one!

PR #1508, opened 2026-08-05
fix(): Cal50 fix for flip

### LINKED ISSUE #1374: Cal50 ver. 1.01 , glitces when running on 4k oleds, maybe issue with scaling
## Caliber 50

I starts out ok, then on the main screens it's missing the forest background. 
And when you start the game it start to glitch. 

Maybe it's a mister.ini setting attached the .ini below
( I would actually like the scaling not to go fullscreen - I don't mind borders as long as the aspect ratio is corret )
I have tried 1920x1080, 1920x1200 and 2560x1440 both 60 and 50 hz  

--- 
 
--- 

--- 

--- 

----

### My mister.ini
https://pastebin.com/3H5uCX98

asturur:
closes #1374 

but also just a draft


jotego:
I have to tested but it looks reasonable. Did you test it?

asturur:
not yet is a draft i m still looking. It renders ok in sim, but don't look at this stage is wasted time i mark it ready for review when i think it is.

asturur:
@jotego when you added `OBJ_LIMIT(9'd200)` in the cal50 commit in master, did you had some reason backing it? Using this lower limit breaks the title screen.

image

The missing yellow band on the title screen is sprites. putting back the limit to 1ff bring it back, but i wonder why you lowered.

asturur:
Also i checked mame to decide which text has the correct alignment, and i mame is fully broken with flip in this game. So this could be something else to contribute back

asturur:
this game still renders ar 368px

cores/cal50/cfg/mame2mra.toml
```diff
@@ -19,12 +19,9 @@ skip.Machines=[
 
 [buttons]
 names = [
-    { machine="calibr50", names="Shoot,Grenade,Turn Left,Turn Right" },
+    { machine="calibr50", names="Shoot,Grenade,Turn Right,Turn Left" },
```
asturur, line 22: verified those are inverted in game

cores/cal50/cfg/mame2mra.toml
```diff
@@ -19,12 +19,9 @@ skip.Machines=[
...
 ]
 
-[header]
-frames = [ { width=8 }, ]
```
asturur, line 26: restore full 384 game pixels

asturur:
I played this a bit, it seems ok. the only issue is if we are ok with the changes.

PR #1510, opened 2026-08-06
fix(): Proposed a fix for chuka taisen issue on kiwi core.

### LINKED ISSUE #1509: fix(): Support chuka taisen in the kiwi core
> Can the jtkiwi core be revisited at some point? Chuka Taisen (=Cloud Master) and Plump Pop are still missing and I guess that a bit of Taito knowledge was gained in the meantime by working on the Fairyland Story. 

 _Originally posted by @real-amano in [#710](https://github.com/jotego/jtcores/issues/710#issuecomment-3232816567)_

I investigated this a bit with taps into simulations.
Cloud master has a sprite table of 512 like the other games of this core.
At boot it initialize only a portion of them leaving 130 sprites at 0.
This means:
position x : 0, position y: 0, sprite code 0.
Now the sprite 0 in chukatai rom is a grey block, and so the first 2 lines as soon as they start are hammered with 130 writes of grey overlapped they eat all the time we have, the next line start and the rest of the line doesn't get drawn.

those 2 lines ( 3 for now ) are because the sprite at 0 is entering 2 lines of the playfield.

What is eating our time is a mix of fetching/writing and i can't really fix much.

So this a mix of issues, even if i would blank the tile 0 with transparent, i would remove the grey block but still i would spend time fetching and possibly drawing those sprites.

We can't avoid rendering sprites for the top 2 lines because the game does render sprites in some death circumstances, it would be an hack worse than the program rom change

I can't stop the sprite limit to the correct size because that would be incorrect too, it would require a new core because the other cores do use 512 sprites and i can't limit to less just because one game has an issue.

we could patch the program rom to initialize all the sprite table with the parked value, that could be a single rom patch at boot but would alter the original game ( is a stretch, is probably a LOOP and i would just increase the value of the loop index)

This is a comparison with mame:

Image

we draw sprites one line below compared to mame but we don't care because we don't really have a way to see which one is the correct one, but also mame has this spurious line and with the infinity bandwidth of the mame this doesn't create issues. A video from real hardware would really help to understand if the grey line is there, can't find any.

DottorToppel, same hardware, same issue, but it operates with flip configuration and in that case the sprites land off screen, no issues.

The combination of non initialized sprites and not flip is what make the bug appear.
Does it happen on original hardware too? i have no idea i can't find videos.

asturur:
The problem is well descripted in the issue #1509 i won't replicate here.

The fix i propose is a patch, not a large hdl improvement on fetch speed that could be hard to chase.

I decided to propose a patch to the program rom that work around the issue because the issue is really a combination of factors that make our sdram access a problem.

We can try to implement a better approach too with hdl that makes the issue less problematic and leave the grey bar there, but this seems reasonable to me.

closes #1509


cores/kiwi/hdl/jtkiwi_obj.v
```diff
@@ -172,7 +174,7 @@ jtframe_obj_buffer #(
     .flip   ( 1'b0      ),
     .LHBL   ( ~hs       ),
     // New line writting
-    .we     ( buf_we    ),
+    .we     ( buf_we & ~hs ),
```
asturur: @jotego this is avoiding spilling sprites from line N to line N + 1 when there is a line time issue. Chuka taisen weas making 3 lines dirt because line 1 was spilling to line 2, and line 2 to line 3. Now with the rom patch this isn't an issue anymore but still fills like something that in general is correct. But i m also worried it could affect new sprites at the start of the line.
It depends when the code that executes the sprite drawing start drawing and when is the correct time for line N and line N + 1.

I would leave a note about this change but not chase it now, unless you think it makes sense.

jotego:
This is a bit of a hack but it doesn't hide that fact so we can fix it better in the future. I think it is ok. 

@skutis did you RE'ed these chips? I am not sure if you had looked into this...

skutis:
I have not RE:d the chips yet. I have die pictures though.

asturur:
yes this is a rom patch waiting for full decap analysis.

An avenue i didn't test, what if this game was default rotated by 180 and flipped? That would make it lik drtoppel and working fine no patches

asturur:
no sorry rotation + flip would make the game go on the left rather than on the right.
If this game was default flipY on, and installed on a cab with the mirror, that would explain.

asturur:
i ll test this and then mark for review.

asturur:
Confirmed working good.

asturur:
Someone that tested the core compared the audio to mame and noticed this core sounds bad:

// core
https://github.com/user-attachments/assets/1bd9996b-26f5-48c1-ae6e-2735c36bd191

// mame
https://github.com/user-attachments/assets/0bc2bd0c-0b9c-4cf9-aae1-08d0956623a8

asturur:
Ok i noticed that chuka taisen was getting the FM chip clocked to 1.5Mhz but also when switched to 3Mhz it sounds terrible.

I tried remove lumping, that was something that was causing issues also on double dribble and the sound fixed.
This would alter all the other games so i didn't commit and push it ( is in a separate branch ) but here you can hear the differences.

[fpga_nolump_fm3.wav](https://github.com/user-attachments/files/30864941/fpga_nolump.wav)
[fpga_nolump_fm15.wav](https://github.com/user-attachments/files/30864942/fpga_nolump15.wav)
[mame_chukatai.wav](https://github.com/user-attachments/files/30864943/mame_chukatai.wav)
[fpga_fm3.wav](https://github.com/user-attachments/files/30864947/fpga_fm3.wav)

asturur:
without lumping it sounds good. But for now that is a different topic, since all the other games needs to be tested or lumped manually with an header bit.

asturur:
this is the game with audio NOT LUMPED.

https://github.com/user-attachments/assets/fa4832fe-7036-4193-9543-08cb945dea18

PR #1511, opened 2026-08-06
perf(kiwi): JUST A TEST overlap object scan with drawing, cache obj fetches

asturur:
THIS IS JUST A TEST TO BUILD TO SEE IF DIFFERENT HANDLING OF SPRITE WRITES HELPS WITH CHUKATAI. DO NOT REVIEW.


jotego:
jtutil has a command to request builds from the terminal. You can use that instead of a PR. Have a look.

asturur:
i can also just make a branch and build from the action that is what i do, but i wanted to see how the diff looked like.

Issue #1481, opened 2026-07-18
SiDi128 Screen Rotation broken


danyppc:
Screen rotation broken with yesterday 18 Jul 2026 on SiDi128

jotego:
Glad to see that someone is using the SiDi128! I'll check it

danyppc:
I own 3x SiDi128 (1 is of my brother), 2x MiSTica, 3x MiST 
:)

rp-jt:
@jotego Could this be related to commit 3bf8a4ba3c642c5a4ebde6a81e49363366ffcc59 ? 
In this, `modules/jtframe/target/mist/hdl/jtframe_mist_top.sv` changes an `endif` for an `else` in line 144. So now we have this:
```
`ifdef MIST_DUAL_SDRAM

`ifdef JTFRAME_LF_SDRAM_BUFFER
`define MIST_USE_SDRAM2
`else
`ifdef JTFRAME_VERTICAL
`ifdef JTFRAME_SDRAM_ROTATION
`define MIST_USE_SDRAM2
`define MIST_USE_SDRAM2_ROTATION
`endif
`endif
`endif
```
 In the former version,  `MIST_USE_SDRAM2_ROTATION` is defined if `JTFRAME_VERTICAL` and `JTFRAME_SDRAM_ROTATION` are defined, independently from `JTFRAME_LF_SDRAM_BUFFER`, but now, both Rotation and frame buffer are incompatible.

Currently, `MIST_USE_SDRAM2_ROTATION` cannot be defined in SiDi128, because both `MIST_DUAL_SDRAM=1` and `JTFRAME_LF_SDRAM_BUFFER=1` are found in `modules/jtframe/target/sidi128/target.def`

rp-jt:
Frame buffer should take priority over vertical video for `gals` to work, so I've tried using this block instead:
```
`ifdef MIST_DUAL_SDRAM

`ifdef JTFRAME_LF_BUFFER
`ifdef JTFRAME_LF_SDRAM_BUFFER
`define MIST_USE_SDRAM2
`endif
`else
`ifdef JTFRAME_VERTICAL
`ifdef JTFRAME_SDRAM_ROTATION
`define MIST_USE_SDRAM2
`define MIST_USE_SDRAM2_ROTATION
`endif
`endif
`endif
```

However, the resulting compilation showed no change when selecting `Screen Rotation: YES` in the sidi128 OSD menu. I was testing `1942` and `Rumba Lumber`
[...]

PR #1517, opened 2026-08-10
SiDi128: fix rotation (#1481)

gyurco:
BUG: #1481


jotego:
I just saw this PR, right after I had pushed an edit which is virtually identical.

Thanks for taking the time to make the PR!

PR #1523, opened 2026-08-17
ascal: Replace output handshake with gray counter

littlegamer87:
This PR fixes a deadlock in ascal that can occur with `vsync_adjust=2` and video mode changes. The counter at the scaler output will break if 2 acks are missing. Solve by replacing the req-ack handshake with a gray counter that can tolerate the jumpy `o_clk`.

This patch is accepted for the mister template. There is one PR still open to make the counters smaller, I overdid it a little. That correction is already included in this PR.


littlegamer87:
Already superseded, sorry. Closing

PR #1527, opened 2026-08-18
JTDOCASTLE (Universal Do's Castle family, 9 games)

meathax:
Adds `cores/docastle`: Universal's Do's Castle board family — nine games
across three hardware profiles (Castle / Run Run / Soccer) on one core.
Three Z80s, 4x SN76489A, HD6845S CRTC, MSM5205 ADPCM on Soccer, and the
custom CF37201N sprite chip.

Built from the Universal service manual schematics and furrtek's CF37201N
(TAL004) decap work; MAME's universal/docastle.cpp used as cross-check only.

Status — passing:
- jtframe cfgstr / mem / files / msg / mra
- Verilator lint, top jtdocastle_game_sdram: 0 errors, 0 warnings
- 9 MRAs generate; all 123 ROM CRCs verified against an independent reference

Known gaps:

1. gfx1_ok / gfx2_ok are ignored, *_cs tied high. Per jtframe_romrq.v that's
   only valid while the address doesn't change before data_ok, which the
   sprite state machine breaks. A bare `ST_GWAIT: if(gfx2_ok)` isn't enough
   either — OKLATCH=1 keeps data_ok high a cycle after the address changes.
   Full analysis in jtdocastle_game.v. Blocker for hardware.
2. No ver/ scenes, so unit_simulations has nothing to run. Guidance on the
   expected scene format for a new core would help.
3. doc/mame.xml had no docastle.cpp entries, so 20 machines were merged in
   from a local `mame -listxml`. That file is stamped 0.282 while these came
   from 0.289 — ROM CRCs are identical, but you may prefer to regenerate it
   with `jtframe mra --reduce`.

jt89 is carried core-locally
under hdl/ as there's no modules/jt89 submodule.


jotego:
This is quite crazy indeed. But well, I was expecting something like that. I cannot even review the PCB against it as it is very convoluted right now. Let's start by cleaning the code in 2-3 rounds so we can actually see what is going underneath all the layers.

The known gaps shown in the PR comment are all requirements. We can accept not having scene simulations but the other two points should be fixed. Especially expanding doc/mame.xml correctly and letting that hold the information for the same MAME version as the rest of the entries.

Other than that:

## Changes requested 

1. The CRTC timing is wrong: `pxl_cen` is 6.143465 MHz with `JTFRAME_PXLCLK=6`, so the 312×264 raster runs at about **74.59 Hz**, not 59.659 Hz. It looks like the game can program the video timing? If that's the case maybe this automatic review item is not doing the right thing. If there is a programmable chip indeed in jtdocastle_crtc, then it might benefit from having its own cfg/mmr.yaml configuration there.

2. Graphics and ADPCM SDRAM requests ignore `*_ok`, while addresses change. `gfx1_cs`/`gfx2_cs` are permanently high and the sprite renderer assumes one-cycle latency, so stale ROM bytes can be consumed. This is an acknowledged hardware blocker in the PR. [game](https://github.com/jotego/jtcores/blob/b5536c31c0efe65399702e47ec2cdbea1564d53b/cores/docastle/hdl/jtdocastle_game.v#L857) · [renderer](https://github.com/jotego/jtcores/blob/b5536c31c0efe65399702e47ec2cdbea1564d53b/cores/docastle/hdl/jtdocastle_video.v#L353)

3. The required Z80 wait paths are deliberately unwired: main/sub mailbox wait and PSG READY wait. Both can cause lost or mistimed board transactions. [main](https://github.com/jotego/jtcores/blob/b5536c31c0efe65399702e47ec2cdbea1564d53b/cores/docastle/hdl/jtdocastle_main.v#L124) · [sub](https://github.com/jotego/jtcores/blob/b5536c31c0efe65399702e47ec2cdbea1564d53b/cores/docastle/hdl/jtdocastle_sub.v#L113)

4. Move CPU-visible tile, colour, and sprite RAM into `cfg/mem.yaml` BRAM definitions, with scene save/restore. They are currently ad-hoc arrays in the video module. [video](https://github.com/jotego/jtcores/blob/b5536c31c0efe65399702e47ec2cdbea1564d53b/cores/docastle/hdl/jtdocastle_video.v#L146)

5. Implement `gfx_en` layer gating in colmix; it is currently accepted but unused. [colmix](https://github.com/jotego/jtcores/blob/b5536c31c0efe65399702e47ec2cdbea1564d53b/cores/docastle/hdl/jtdocastle_colmix.v#L117)

6. This one is optional: Add real validation: saved scenes, boot-to-credit frame, and coin/audio CAB test for representative variants. `reg.yaml` references nine `reg.cab` files that are absent. 

7. Fix the `jt5205` file-list warning, trailing whitespace in vendored `jt12`/`jt89` files, use a two-decimal `JTFRAME_RATE`

8. Remove all duplicated jt12 and jt89 files and instantiate them from the jt12 submodule instead
9. The corename must be six-character long for compatibility with Analogue Pocket. Make it something like "dcastl" 
10. Remove all long comments from the code. The code should be self-explaining for an expert
11. Remove all explicit references to Quartus or intel specific elements like `ramstyle`. Instantiate memory seen by both the CPU and the video modules in mem.yaml

meathax:
All 11 items addressed, in rounds as you suggested. Summary:

1. **CRTC timing** — real bug: `JTFRAME_PXLCLK=6` made `pxl_cen` a
   framework *input* at 6.14 MHz. Dropped; `pxl_cen`/`pxl2_cen` are now
   outputs driven from one mem.yaml entry giving MCLK (9.828 MHz) and
   MCLK/2 (4.914 MHz), the board's real relationship. And yes — the
   HD6845 is software-programmed (R0–R17 via the indexed port pair), so
   a fixed-raster refresh calc doesn't describe this board. Register
   file now generated from `cfg/mmr.yaml`; the frame-shadow commit is
   kept, since that's what stops a write burst tearing the raster.
2. **SDRAM `*_ok`** — two different problems. The tile fetch is
   combinational per pixel and can't stall, so gfx1 (fixed 16 kB) moved
   to BRAM. The sprite FSM now uses the per-request `cs` toggle
   `jtframe_romrq` documents. ADPCM gained `rom_cs`/`rom_ok`.
3. **Z80 waits** — both wired. Mailbox wait and PSG READY each suppress
   their CPU's clock enable, same as `jtframe_z80wait`'s `dev_busy`.
4. **Shared RAM** — tile, colour *and* sprite RAM are all mem.yaml
   dual-port BRAM with scene save/restore. The sprite scanner rebuilds
   entries byte-wise (8 clocks/sprite; a line has 3120).
5. **`gfx_en`** — implemented, gating pens before the priority mux so a
   disabled layer stops winning priority.
6. Scenes not added. But `reg.yaml` no longer references nine absent
   `reg.cab` files — entries are bare, like castle/circus.
7. No jt5205 warning, no trailing whitespace, `JTFRAME_RATE=59.66`.
8. All 9 duplicated jt12/jt89 files removed, pulled from the submodule
   via `jt89.yaml`. `jt89_sms.v`/`jt12_dac2.v` were unused.
9. `JTDCASTL` / `cores/dcastl` / `jtdcastl_*`.
10. Comments 1172 → 90 lines (33% → 3.6%; flstory is 4%). Verified
    behaviour-neutral: comment-stripped source hashes identical.
11. No `ramstyle` anywhere.

Also `doc/mame.xml`: `docastle.cpp` only changed between 0.282 and 0.289
in the palette function (blue normalised to 0x52/0xad) and blank lines,
neither of which reaches listxml — so the entries are equivalent. The
one real difference was `device_ref` gaining a `tag` attribute in 0.289;
stripped, so they're now element-for-element like the rest of the file.
That palette commit also confirms the blue weights this core implements.

Gates: `jtframe cfgstr/mem/files/msg/mmr/mra` all pass, Verilator lint
0 errors / 0 warnings.

happy to take direction on anything you'd shape
differently, particularly the sprite scanner's fetch timing.

jotego:
**THE TEXT BELOW IS AN AUTOMATIC REVIEW. DO NOT DO ANYTHING YET. I WANT TO DO A MANUAL REVIEW LATER TODAY**

### **Automatic Review**

Much improved: timing, SDRAM-ready handling, CPU waits, default video BRAM, and CRTC MMR definition were addressed. I still request changes before merging:

1. `gfx_en[0]` is wrong for Run Run/Soccer. Disabling tiles forces `tile_pen_en` to zero, but with `low_pen_priority=1` that makes `tile_front=1`, so sprites remain hidden instead of isolated. Gate `tile_front` itself with `gfx_en[0]`. [colmix](https://github.com/jotego/jtcores/blob/1746845f606949d9119ee4c217446f46b137a4d2/cores/dcastl/hdl/jtdcastl_colmix.v#L48)

2. Scene IOCTL support is incomplete. `JTFRAME_IOCTL_RD` exposes `ioctl_din`, but it is undriven (confirmed by Verilator). The CRTC MMR also ties its IOCTL address to zero and discards its dump output, so its CPU-visible state cannot be saved/restored with a scene. Route and mux those signals. [CRTC](https://github.com/jotego/jtcores/blob/1746845f606949d9119ee4c217446f46b137a4d2/cores/dcastl/hdl/jtdcastl_crtc.v#L55) · [memory config](https://github.com/jotego/jtcores/blob/1746845f606949d9119ee4c217446f46b137a4d2/cores/dcastl/cfg/mem.yaml#L93)

3. Align the framework `sample` strobe with `cen_48k`. It is currently a `/1000` divider of the 49.153 MHz clock (~49.15 kHz), while the audio filter runs at `cen_48k`; that is a 2.4% mismatch. [game](https://github.com/jotego/jtcores/blob/1746845f606949d9119ee4c217446f46b137a4d2/cores/dcastl/hdl/jtdcastl_game.v#L503)

Validation status:

- `jtframe mra dcastl` generated all nine set MRAs.
- `git diff --check` passes.
- `lint-one.sh dcastl` and its `NOMAIN`/`NOSOUND` variant run, but both emit the undriven-`ioctl_din` warning.
- No scenes/CAB scripts, boot-to-credit capture, coin-audio test, successful Quartus build, hardware test, or GitHub CI result is available.

jotego:
additional action items:

- **P0: Sprite CPU’s first copy phase has no effect.** `copy_addr`, `copy_data`, and `copy_we` are only routed into the game module and then placed in `_unused`; the displayed sprite RAM is written directly from the main CPU. The README claims both sprite-CPU bus phases are implemented, but this phase cannot affect rendering. Wire it to the intended shared sprite RAM (with proper arbitration), or remove the unsupported claim/path. [game wiring](https://github.com/jotego/jtcores/blob/1746845f606949d9119ee4c217446f46b137a4d2/cores/dcastl/hdl/jtdcastl_game.v#L166-L180) · [unused outputs](https://github.com/jotego/jtcores/blob/1746845f606949d9119ee4c217446f46b137a4d2/cores/dcastl/hdl/jtdcastl_game.v#L539-L549) · [copy generation](https://github.com/jotego/jtcores/blob/1746845f606949d9119ee4c217446f46b137a4d2/cores/dcastl/hdl/jtdcastl_spritecpu.v#L109-L128)

- **P1: Remove the duplicated, disconnected RGB/PROM mixer from `jtdcastl_video`.** The game leaves `prom_addr`, `r`, `g`, and `b` unconnected, while `jtdcastl_colmix` is the actual mixer. Keeping a second priority/PROM implementation in video invites divergence—especially since `gfx_en` is applied only in colmix. Video should emit layer taps; colmix should exclusively decide priority and DAC output. [disconnected ports](https://github.com/jotego/jtcores/blob/1746845f606949d9119ee4c217446f46b137a4d2/cores/dcastl/hdl/jtdcastl_game.v#L396-L412) · [dead mixer](https://github.com/jotego/jtcores/blob/1746845f606949d9119ee4c217446f46b137a4d2/cores/dcastl/hdl/jtdcastl_video.v#L360-L375)

- **P1: Add real `NOMAIN`/`NOSOUND` modes before scene work.** There are no such guards in the tracked source, so there is no controlled video-only/audio-less simulation configuration. Add them to the CPU and sound hierarchy, then add at least one boot/coin scene. This also makes the existing unverified renderer easier to isolate.

- **P2: Do not ship the inactive PCB renderer as normal production RTL.** It is instantiated even though `PCB_FRAMEBUFFER=0`, contains another module in the same file, and uses bespoke inferred line/field RAM machinery. Either move it to an explicitly experimental/tested implementation or refactor it around JTFRAME sprite/RAM helpers. [instance](https://github.com/jotego/jtcores/blob/1746845f606949d9119ee4c217446f46b137a4d2/cores/dcastl/hdl/jtdcastl_video.v#L218-L237)

- **P2: Restrict `JTFRAME_IOCTL_RD` to debug until the complete scene path is wired.** This is not NVRAM, yet it is enabled globally in `macros.def`, creating a production `ioctl_din` port that remains undriven. Put it behind the debug configuration after fixing the MMR/RAM dump mux. [macro](https://github.com/jotego/jtcores/blob/1746845f606949d9119ee4c217446f46b137a4d2/cores/dcastl/cfg/macros.def#L19-L22)

- **P3: Clean up source-contract debt.** The new RTL headers use `Author: aCORES` rather than the project’s required author, timing comments still describe a local `/10` divider no longer present, and several comments reference nonexistent review items such as “C3” and “V1”.

Validation: current PR head remains `1746845`; `lint-one.sh dcastl -d NOMAIN -d NOSOUND` still reports the undriven `ioctl_din`.

I used the core-review rubric for integration/testability and the Verilog-module rubric for ownership, module structure, and RTL hygiene.

jotego:
- **Refactor every CPU address decoder into one explicit `always @*` block.** `main`, `sub`/sound, and sprite CPU currently define each `*_cs` as scattered continuous assignments, with profile conditionals and peripheral addresses split across unrelated logic. That makes the map difficult to audit and violates the core rules. Each CPU needs one decoder block that defaults every select low, then decodes the map by profile. [main decoder](https://github.com/jotego/jtcores/blob/1746845f606949d9119ee4c217446f46b137a4d2/cores/dcastl/hdl/jtdcastl_main.v#L125-L152) · [sound decoder](https://github.com/jotego/jtcores/blob/1746845f606949d9119ee4c217446f46b137a4d2/cores/dcastl/hdl/jtdcastl_sub.v#L118-L123) · [sprite CPU decoder](https://github.com/jotego/jtcores/blob/1746845f606949d9119ee4c217446f46b137a4d2/cores/dcastl/hdl/jtdcastl_spritecpu.v#L100-L107)

- **Gate memory selects with a real Z80 memory access.** The current `rom_cs`, `ram_cs`, tile/color/object RAM selects are address-only; they can assert during I/O and refresh cycles. Define, for example, `cpu_macc = ~mreq_n & rfsh_n`, then make memory selects depend on it. Keep I/O decoding separately qualified by `~iorq_n`. Otherwise an I/O access such as the CRTC ports can still generate a ROM request/wait.

- **Decode peripheral selects, then consume them; do not re-decode addresses in sequential blocks.** Main’s CRTC, NMI, watchdog, ADPCM, communication, video, colour, and sprite selects should all come from the main decoder. Sound needs explicit `psg_cs[3:0]`, input, mailbox, ROM, and RAM selects. Sprite CPU needs explicit staging/copy/CF selects. The current PSG address `case`, CRTC I/O `case`, and raw copy-window comparison distribute decode responsibility across the module. [main CRTC special case](https://github.com/jotego/jtcores/blob/1746845f606949d9119ee4c217446f46b137a4d2/cores/dcastl/hdl/jtdcastl_main.v#L196-L207) · [PSG special case](https://github.com/jotego/jtcores/blob/1746845f606949d9119ee4c217446f46b137a4d2/cores/dcastl/hdl/jtdcastl_sub.v#L191-L209)

- **Use standard bus strobes and edge helpers.** Define `cpu_rd`, `cpu_wr`, `cpu_iord`, and `cpu_iowr` once from the qualified bus cycles. Generate one-cycle peripheral strobes from those signals, using `jtframe_edge` where an event/acknowledge edge is required. The hand-rolled `wr_d`, `rd_d`, and `io_wr_d` blocks are duplicated and make side-effect timing hard to verify.

- **Use the required CPU-module layout.** Declarations → decoder → input-data mux → peripheral state → CPU instance last. Every CPU is currently instantiated before its decoder and peripheral logic. This hides the memory map above the CPU rather than presenting it as the module’s primary contract.

- **Use a ternary `cpu_din` mux.** Replace the sequential if/else ladders with the aligned priority mux convention. The sound input selector should be a registered cabinet-data mux, not a single-use function plus extra aliases. [main mux](https://github.com/jotego/jtcores/blob/1746845f606949d9119ee4c217446f46b137a4d2/cores/dcastl/hdl/jtdcastl_main.v#L170-L178) · [sound mux/function](https://github.com/jotego/jtcores/blob/1746845f606949d9119ee4c217446f46b137a4d2/cores/dcastl/hdl/jtdcastl_sub.v#L125-L176)

- **Make profile-dependent maps auditable.** The three profile maps belong as adjacent cases in the main decoder, with a documented table copied from MAME/schematic evidence in `doc/`. The current nested ternaries make overlap and holes difficult to see, especially around ROM/RAM/sprite ranges.

- **Put the shared sprite staging RAM behind an explicit decoder and arbitration boundary.** It is simultaneously main-written and sprite-CPU-read, but the main-side access is injected as a side port rather than represented by a shared-memory contract. That prevents a review of ownership, CPU timing, and scene restore.

These CPU-decoder changes belong together with the architecture refactor: `jtdcastl_main`, `jtdcastl_sound`, and `jtdcastl_spritecpu` should each have one readable, evidence-backed bus map, rather than exposing address fragments upward into `game`.

jotego:
`mame2mra.toml` has a release-blocking coverage defect plus several structural issues.

- **P0 — Only six of the advertised nine games ship.** Regenerating the MRAs leaves `release/mra` with six files. The selected clones `douni`, `spiero`, and `asoccer` are omitted because `[parse]` has no `main_setnames`. Add:

  ```toml
  [parse]
  main_setnames=["douni","spiero","asoccer"]
  ```

  This is required if the README’s “nine games” claim is intended to mean nine released MRAs.

- **P0 — Explicitly exclude or support every other official set in `docastle.cpp`.** The configuration only skips bootlegs, but MAME also has `docastle2`, `docastleo`, `dorunrun2`, `dorunrunc`, `dorunrunca`, `idsoccera`, and `idsoccert`. None has header/profile data. The comment claiming sourcefile grouping safely handles clones is false: with a fuller MAME XML, those can yield MRAs filled with header `ff`, leaving the core in reset. Add them to `skip.setnames`, or add verified profiles and header entries.

- **P1 — Replace `[header].data` with `[header].registers`.** The game ID is a configuration register, but the TOML emits raw bytes and `jtdcastl_game` manually captures the download stream. Define a generated `game_id` register at `0[3:0]`, with the nine set values, instantiate `jtdcastl_header`, and remove manual header capture. This is specifically required by the core-review rules. [header section](https://github.com/jotego/jtcores/blob/1746845f606949d9119ee4c217446f46b137a4d2/cores/dcastl/cfg/mame2mra.toml#L35-L82)

- **P1 — Rework the graphics ROM width/layout together with the video refactor.** The TOML declares every region as 8-bit, which hard-couples `gfx1`/`gfx2` to the bespoke byte-at-a-time fix/object renderers. The replacement `jtframe_tilemap` and `jtframe_objdraw` paths need helper-compatible packed ROM words—normally a wider region/bus and verified `sequence`/packing. [8-bit-layout claim](https://github.com/jotego/jtcores/blob/1746845f606949d9119ee4c217446f46b137a4d2/cores/dcastl/cfg/mame2mra.toml#L101-L126)

- **P2 — Correct the ROM-layout documentation.** The comments say `ba1=gfx1`, but the actual region table places `gfx2` at `JTFRAME_BA1_START`; `gfx1` is BRAM at `JTFRAME_PROM_START`. The same stale bank narrative exists in `macros.def`. This must match `mem.yaml` before the layout can be audited.

- **P2 — Set proper MRA ownership metadata.** `[global].author=["aCORES"]` is not acceptable project provenance; use the proper author/maintainer attribution.

The current region order and explicit start offsets are internally consistent with the present `mem.yaml`, and `jtframe mra dcastl` completes successfully. The generation result, however, confirms only six released MRAs, so the multi-game configuration is not complete.

meathax:
All items from this round are addressed in `1ce302b1e` and `24a096453`.

**The 3 direct bugs**
- `gfx_en[0]` priority bug: `tile_front` now derives from raw `tile_pen[3]`, not the gated `tile_pen_en[3]`.
- Scene IOCTL dump: CRTC MMR's `ioctl_addr`/`ioctl_din` are now actually wired end-to-end (crtc → video → game), gated by `JTFRAME_IOCTL_RD`.
- `sample` strobe: was a local ~49.15kHz `/1000` divider, now aliased to `cen_48k` (the audio filter's own enable).

**P0/P1/P2/P3 items**
- Sprite CPU's first-phase (`0x8000` staging RAM) copy path: verified against mrdo's `docastle_spritecpu.sv`/`docastle_core.sv` ground truth — on real hardware it only ever fed a debug tap, never the displayed sprite output. Removed the dead ports; README corrected.
- Removed the dead PROM/RGB mixer in video.v (duplicated jtdcastl_colmix.v, confirmed unconnected).
- NOMAIN/NOSOUND guard added to jtdcastl_sub.v, following jt1942_sound.v.
- jtdcastl_pcb_sprite.v split into one module per file (+ new jtdcastl_field_ram.v).
- `JTFRAME_IOCTL_RD` moved behind the `debug` macro-line convention.
- `Author:` fields corrected to the real git-log committer identity across all touched files.

**CPU decoder refactor**
main/sub/spritecpu all rewritten to the jtajax_sub.v house style: one explicit `always @*` decoder, `cpu_macc` gating, `jtframe_edge`/`jtframe_edge_pulse` only for genuine one-shot events, ternary `cpu_din` mux, CPU instance last. New `doc/main_map.md` documents the profile-dependent address map. (Caught and fixed a duplicate `rom_cs` declaration this introduced in jtdcastl_sub.v via Verilator lint.)

**mame2mra.toml**
- 6/9 → 9/9 MRAs now generate (`main_setnames` for douni/spiero/asoccer, which are clones in mame.xml).
- `skip.Setnames` added for the 7 undocumented clone sets.
- `[header].data` → `.registers`, generating `jtdcastl_header.v`; `game_id_ok` kept as an explicit sibling register in jtdcastl_game.v so the `fill=0xff` reset-hold safety property survives unchanged.
- **Bug I found in the process, not on your list**: `machine=` in jtframe's `Selectable.Match()` only self-matches a set with `Cloneof==""`. douni/spiero/asoccer (all clones) never matched their own `machine=` entry in the registers table and were silently inheriting their *parent's* game_id instead of their own. Fixed with `setname=` for those three. Verified: all 9 MRAs' header nibble now reads correctly (0x0–0x8), and every ROM part's CRC32 cross-checked clean against an independently-validated reference set.

**jtframe_tilemap / jtframe_objdraw — investigated, deliberately not adopted**
Checked jtframe_tilemap.v and jtframe_draw's ROM contract directly: both require bit-planar pixel data. dcastl's real, decap-verified gfx ROMs are nibble-packed (2 pixels/byte) — a genuine physical format mismatch, not a naming one. Precedent for exactly this situation already exists in-tree: `cores/kicker/hdl/jtkicker_objdraw.v` (used by `cores/circus`) doesn't use the generic module either, for the same reason.

I also checked whether just the double-buffer primitive (`jtframe_obj_buffer`) could drop in under dcastl's own sprite scan logic. It's built on `jtframe_dual_ram`, a registered/synchronous-read BRAM, whereas dcastl's current line buffers are combinational-read — swapping would add real read latency and could shift sprite pixels horizontally by an amount I have no way to verify here (no MAME/Verilator pixel-diff harness exists yet for this core in jtcores). Given the renderer is already decap-verified and has been through working RBF builds, I didn't want to trade that for an unverified timing change. Documented the full reasoning in jtdcastl_video.v (above `line0_even`) and in the README's Known Limitations. Happy to revisit if you'd rather I build out that verification harness first, or if you see a reason the latency concern doesn't actually apply here.

Verification this round: `jtframe cfgstr/mem/mmr/mra` all exit 0, 9/9 MRAs generate, Verilator `--lint-only` is 0 errors/0 warnings on the full file list.

meathax:
Apologies if it feels like your are having a conversation with an AI (yes i watched your interview today), I am reading and learning a lot and using it to enhance my workflow, appreciate your help!

jotego:
We have to give up on trying to convert this code. It is too convoluted. We will work together on something else. Thanks for your time.

PR #1530, opened 2026-08-21
Space Harrier (Rewrite)

niknakniknak:
Redone from schematics, JTFrame, my Arcade Mini DX and boards. Mame. in that order.

Note, the PCM will sound distorted, and will occasionally drop a voice. Fixes for both exist and work, but JT asked to see the PCM with the problem present, so they are left out of this PR. jtoutrun_pcm is instantiated without .WD(16), and jtoutrun_pcm.v is stock. Both are one-line changes.

60-80 MP, high-resolution composite mode PCB shots: https://github.com/niknakniknak/Space-Harrier-PCBs


jotego:
Please switch to creating the branches in jotego/jtcores so the checks can run.

niknakniknak:
moved to #1564

PR #1542, opened 2026-08-23
Gradius III Scoreboard Fix

niknakniknak:
Before:
before
After: 
after


jotego:
Tested on several games and compared with the baseline:

- Crime Fighters: 597 changed frames
- GBusters: 37 changed frames
- ThunderX: 126 changed frames, plus one missing changed-frame dump
- Aliens and Super Contra: no differences in these captures
- Riders and Simpsons-family games: pixel-identical, as expected because they use `ZENLARGE=1`

The changed frames are usually a one pixel shift of some sprites to the right, but it actually helps. At least, it helps on Crime Fighters:

Before
frame_02078-baseline

After
frame_02078-patched

The left pixel of the blue 2 character is missing on the baseline. So I think this PR is good.

@niknakniknak thank you!

PR #1547, opened 2026-08-24
Gradius III Audio Filter Model & Balance

niknakniknak:
This fixes the filter to the correct values from the schematic. The 007232's output ladder is a Konami custom with no published resistance so the FM/PCM balance was matched against an audio recording of the PCB ( https://vgmdb.net/album/325 )


jotego:
The FM is good. The PCM is not. I will take the PR and make a second commit for the PCM.

Issue #1413, opened 2026-05-25
grad3: randomly losing


rp-jt:
Some users report losing in the first stage even though no hit was received.

Since there are usually many objects on screen and projectiles coming from different angles, it is something hard to catch sometimes.

It is not consistent, I have been able to finish the desert stage without this happening or it happening before even the dunes appear. I don't know if this might be related to an input combination at some point. It occurs at different points on the screen as well

jotego:
I think the problem was that there was no separate video DTACK for each CPU, so they could be getting wrong data. Fixed in ce53a08e1

rp-jt:
Users niknak and kammerer report that it still happens in this new version, but not at all as often as before and this one is much more playable

yukky37:
I am still experiencing this issue with the version released on July 5th. 

In my case, it happens during the attract mode (demo loop) of the Plant Stage (Stage 8). While it doesn't happen every time, it seems more likely to occur during the very first demo loop right after booting up the core.

Since the bug still persists in the latest build, I believe this issue should be reopened for further investigation.

jotego:
I found another related weakness in the design and fixed it in b3739ec05. This fix still needs testing, though.

yukky37:
I tested the August 9th build (`Version 1be22f1 - 2026-8-9`), but unfortunately, the issue still persists.

In my case, it occurs in the exact same spot as I reported earlier: during the Plant Stage (Stage 8) attract mode, especially during the very first demo loop right after booting up the core.

Since it still happens on commit `1be22f1`, could you please reopen this issue?

jotego:
An user said that on stage 4 there are many enemies missing. So this looks more like a bug in the object layer rendering, instead of the previous leads around the main-sub CPU communication.

yukky37:
Actually, I don't think simple sprite invisibility explains the deaths in attract mode. 

Attract mode plays back a fixed sequence of inputs. If the enemies were merely invisible while their logic and hitboxes remained unchanged, the demo would still play out correctly without dying. 

Since the demo loop ends up dying, it suggests that the internal game logic, spawn timing, or processing sync itself is breaking, rather than just a visual rendering issue.

real-amano:
Congratulations, Nik, for finding the root case!

niknakniknak:
Thanks, just a note about the stage 4 issue. I did check that using an invulnerability patch this morning because I am NOT good at this game. it does match the playthroughs so I guess that's fixed too..

PR #1548, opened 2026-08-24
Gradius III Correctness Fix maybe

niknakniknak:
On the sub CPU I noticed all three interrupt clears carry ~VPAn, so acknoledging any one interrupt also wipes the other two. The game's own handlers clear only their own mask bit and deliberately leave the others armed. I can't show this changes anything you can see. Just found it while looking at #1413, it doesn't fix that. Happy to drop it if the ~VPAn is there on purpose?

### DIFF
--- a/cores/grad3/hdl/jtgrad3_sub.v
+++ b/cores/grad3/hdl/jtgrad3_sub.v
@@ -64,9 +64,9 @@ reg  [ 2:0] irq_mask;
 
 assign rst_cpu   = rst | sub_rst;
 assign lvbln     = ~LVBL;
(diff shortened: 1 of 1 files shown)


jotego:
Indeed, the schematics don't do it as the core does. I guess this was carried over from the first AI build and didn't get clean during the schematic review.

A similar fix is need on the main CPU side too. I'll add that one after merging this.

Thank you

jotego:
The PR wasn't still the right fix. Following @skutis schematics, the right implementation is now in 7b735832eaaae6e9d6afc465ab38b1d8b9ee4821.

I am a bit surprised we did not correct this during the initial core review, but well, that's life.

PR #1549, opened 2026-08-24
Raiga Mixer Fix for #1198

### LINKED ISSUE #1198: gaiden: Raiga Strato Fighter title transparency
Noticed the intro titles has a transparency issue with the title write on and the planet showing through the text

Image

Image

niknakniknak:
Fixes #1198 and the worm level

Tested all games on this core and no visual regressions from what I can see. - Ninja Gaiden uses this blend trick too for enemy deaths and looks like the PCB based on YT recordings. Wild Fang doesn't use it.

The board keeps three copies of the palette. The normal colours, the colours a pixel shows when something blended sits in front of it, and the colours the blended thing itself adds. A blended pixel is the last two added together.

The core only ever read the normal copy, and averaged the two colours 50/50. So two thirds of the palette RAM was being written by the game and never read, and the game had no control over how transparent anything was.

With a flat 50/50 the planet came through the letters and the title 'show and hide' band shows through the planet.

Based on mame priority table. Palette Layout from the scene dumps.

### DIFF
--- a/cores/gaiden/hdl/jtgaiden_blender.v
+++ b/cores/gaiden/hdl/jtgaiden_blender.v
@@ -2,7 +2,7 @@
  * SPDX-License-Identifier: GPL-3.0-or-later
  * Date: 2-3-2025 */
 
-// blends two colors by simply averaging the RGB signals
+// blends two colors by adding the RGB signals, with saturation
 module jtgaiden_blender(
     input          clk,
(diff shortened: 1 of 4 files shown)


jotego:
@niknakniknak thank you. In the future, if you create the branches in the repository instead of yours, the checks will run correctly. Thanks!

Issue #1432, opened 2026-06-18
shouse: Blast Off (JP) will not rotate in analog output


Ewkerman:
The setting for flip screen is there but does not seem to do anything. Nothing in Dipswitches either.  It only shows in TATE in one direction (what is upside down for me)

rp-jt:
@jtmiki could you check this? Thanks!

jtmiki:
I have checked it with f8194e6 and the setting for flip screen does not work as Ewkerman said.
Showed up this attached error screen.

<img width="987" height="768" alt="Image" src="https://github.com/user-attachments/assets/a725dfe7-b7ee-4278-b489-1fa7aba33169" />

Ewkerman:
I also found it on 19XX core also.

jotego:
19XX should work

asturur:
I ll have a look with fresh eyes

asturur:
The reason for the error is this one:

```
f549  MCU  WR C100=F6  pc=BDE8   ; MCU writes its test ramp
f549  MCU  WR C101=F7  pc=BDE8   ;   F6, F7, F8
f549  MCU  WR C102=F8  pc=BDE8
f549  SND  RD 7101=F7  pc=D089   ; sound CPU, polling this byte, sees it go non-zero
f549  SND  WR 7101=80  pc=D09A   ; ...and answers by writing 80 over it
f549  SND  RD 7100=F6  pc=D09F
f550  MCU  RD C100=F6  pc=BDF7   ; readback: 0x100 = F6  correct
f550  MCU  RD C101=80  pc=BDF7   ; readback: 0x101 = 80  WRONG, expected F7
                                 ; -> status 02, address C101, error screen
```

the ram test write f6,f7,f8.
meanwhile the sound cpu is monitoring c010, reads f7, put there 80. when we go there to look for F7 we find 80.

asturur:
Would be interesting to know if the same happens on a real board at boot up. this happens only when you enter the dipswitch service menu at bootup. if you go there in mame with the service mode + reboot, it doesn't happen.

asturur:
@Ewkerman how were you able to reach the flip setting? i can't enter in menu at all.

Ewkerman:
I will check tonight.

Ewkerman:
@asturur Confirmed I can not get into Dip Switchs.  Also same issue with the Qix core (if that's yours). 

I thought the video settings would rotate it but it does not on any core via the analog out.

[...]

PR #1555, opened 2026-08-27
doc(): Blastoff memtest issue

asturur:
I summarized in this doc all i could find around this issue that is not easy at all.

All games are affected, but all games can pass the mem test outside blastoff.
This could be a real hardware behavior that would be solved by the warm boot path that instead fails for all games.

I have a workaround for the warm boot issue, but is ugly so i didn't commit it, is the kind of patch that at least written how is written shouldn't fit in a core.

```diff
--- a/cores/shouse/hdl/jtshouse_triram.v
+++ b/cores/shouse/hdl/jtshouse_triram.v
@@ -38,15 +38,21 @@ module jtshouse_triram(

 wire [ 7:0] xdout, xdin, p_bdin;
 wire [10:0] xaddr;
-wire        xwe, bwe;
+wire        xwe, bwe, we1, bcookie_cs, xcookie_cs;
 wire        xsel;
-reg         xsell;
+reg         xsell, rst_l, warm, warm_rd;
+reg  [ 5:0] warm_cnt;
+reg  [ 7:0] flag0, flagf;

 assign xsel  = snd_sel;
 assign xwe   = xsel ? snd_cs & ~srnw : mcu_cs & ~mcu_rnw;
 assign xaddr = xsel ? saddr : mcu_addr;
 assign xdout = xsel ? sdout : mcu_dout;
-assign bwe   = bus_cs & ~brnw;
+assign bcookie_cs = baddr==0 || baddr==11'h00f;
+assign xcookie_cs = xaddr==0 || xaddr==11'h00f;
+assign bwe   = bus_cs & ~brnw & ~rst & ~(warm & bcookie_cs);
+assign we1   = xwe & ~rst & (xaddr != 0 || xdout == 8'ha6 || xsel)
+             & ~(warm & xcookie_cs);

 assign mcu_din = xdin;
 assign bdin = p_bdin;
@@ -88,6 +94,40 @@ always @(posedge clk, negedge srst_n) begin
 end
 `endif

+// Compatibility: preserve the A6/nonzero warm-start cookie until sampled.
+initial begin
+    rst_l    = 0;
+    warm     = 0;
+    warm_rd  = 0;
+    warm_cnt = 0;
+    flag0    = 0;
+    flagf    = 0;
+end
+
+always @(posedge clk) begin
+    rst_l <= rst;
+    if( rst && !rst_l ) begin
+        warm     <= flag0==8'ha6 && flagf!=0;
+        warm_rd  <= 0;
+        warm_cnt <= 0;
+    end else if( !rst ) begin
+        if( warm ) begin
+            if( !warm_rd && bus_cs && brnw && baddr==11'h00f ) begin
+                warm_rd  <= 1;
+                warm_cnt <= 0;
+            end else if( warm_rd ) begin
+                warm_cnt <= warm_cnt+1'd1;
+                // Keep it stable through the BRAM/CPU read latency.
+                if( &warm_cnt ) warm <= 0;
+            end
+        end
+        if( bwe && baddr==0 ) flag0 <= bdout;
+        if( bwe && baddr==11'h00f ) flagf <= bdout;
+        if( we1 && xaddr==0 ) flag0 <= xdout;
+        if( we1 && xaddr==11'h00f ) flagf <= xdout;
+    end
+end
+
 always @(posedge clk, posedge rst) begin
     if( rst ) begin
         snd_din <= 0;
@@ -109,7 +149,7 @@ jtframe_dual_ram #(.AW(11)) u_ram(
     .clk1   ( clk   ),
     .data1  ( xdout ),
     .addr1  ( xaddr ),
-    .we1    ( xwe && (xaddr !=0 || xdout == 8'ha6 || xsel) ), // see https://github.com/jotego/jtcores/issues/410
+    .we1    ( we1 ), // see https://github.com/jotego/jtcores/issues/410
     .q1     ( xdin  )
 );
```

this patch here would be the only one that makes the warm reset path works for every game and the only practical solution i found. The investigation doc/notes should be preserved for future efforts.

related to #1432 

On top of just documenting i proposed a 'fix' i added an imaginary dip switch that would switch the register so that we can flip without using the menu.

An alternative would be embrace the weird patch described in the PR here.

Or dig deeper, find a real board and understand what is going on.


asturur:
This is in draft because i never tried it outside simulation yet.

PR #1556, opened 2026-08-28
Gradius III No-Wait QOL Extra Dip

niknakniknak:
This will enable you to lose a life even quicker.

A No-Wait option to remove (most of) the slowdown that this board is famous for by running the 68000's at 24Mhz instead of 10.
Konami liked to use slow down as a game mechanic feature/not a bug.  They added No-Wait options when these games released on Consoles and this makes pro players happy. Think of it as an extra hardness setting.

This Dip switch setting is defaulted to off - 10Mhz.  Its a live switch so you don't need  to reset the game.


real-amano:
Hmm. This might still come in handy for the 3 level "Beginner Course" in the Japanese Rom.

I guess that the beginner course rules cannot be forced onto the main course. Or the Beginner Course be made to not stop after level III?

niknakniknak:
> Hmm. This might still come in handy for the 3 level "Beginner Course" in the Japanese Rom.
> 
> I guess that the beginner course rules cannot be forced onto the main course. Or the Beginner Course be made to not stop after level III?

It might but it'll only speed things up and make it harder. Not sure what the beginner rules are as I've never touched that after I did many years ago in an arcade and found that the game ended after 3 stages. Felt like I'd been scammed out of my 20p.. but it sounds like that would be more of a rom hack rather than hardware change.

asturur:
can the 68000 actually run at 24mhz with the cen recovery? does it recover ever? with 24Mhz 68000 on a 48Mhz clock has 1 clock of recover each real one.

niknakniknak:
it ends up around 22

jotego:
If keeping it to 12 or 16MHz already gives the boost, I'd rather prefer that. It is a speed that the system can handle consistently and even realistic for the age of the game.

cores/grad3/hdl/jtgrad3_game.v
```diff
@@ -14,6 +14,8 @@ wire        m_tile_dtack, s_tile_dtack;
 wire        tile_irqn, tile_nmin;
 wire        rmrd, prio;
 wire        m_video_req_n, s_video_req_n, m_video_gnt_n, s_video_gnt_n;
+wire        turbo;
+assign turbo      = ~dipsw[20];
```
jotego, line 18: as `dipsw` already goes into main, you can make the turbo assignment there. Not a big deal.

cores/grad3/cfg/mame2mra.toml
```diff
@@ -12,6 +12,9 @@ names = [
 [dipsw]
 bitcnt=20
 delete=[{names=["Unused","Unknown"]}]
+extra=[
+    { name="No Wait", options="On,Off", bits="20" },
```
jotego: Does this get the right default value? It might be on by default.
The name `Turbo` would be easier to understand and consistent with the JTDD/JTDD2 cores as well as with the verilog signal.

cores/grad3/hdl/jtgrad3_main.v
```diff
@@ -241,7 +242,7 @@ jtframe_68kdtack_cen #(.W(6), .RECOVERY(1)) u_dtack(
     .bus_ack    ( 1'b0      ),
     .ASn        ( ASn       ),
     .DSn        ( bus_dsn   ),
-    .num        ( 5'd5      ),
+    .num        ( turbo ? 5'd12 : 5'd5 ),
```
jotego: better to register the static output of muxes in general, although here it is not a big deal.

jotego (changes requested):

niknakniknak:
> If keeping it to 12 or 16MHz already gives the boost, I'd rather prefer that. It is a speed that the system can handle consistently and even realistic for the age of the game.

It doesn't really. I tested 16 & 20 (which were probably a little under that in reality) even 24 doesn't totally stop the most notorious area of slowdown but certainly makes them harder. I did a couple of full run throughs of the game with this at 24. This isn't really a thing from the era - it came with the PS2 version as an extra feature for the hardened players. Will look at the rest shortly.

cores/grad3/cfg/mame2mra.toml
niknakniknak: No wait is Konami's term for this but Turbo is fine.

cores/grad3/hdl/jtgrad3_main.v
niknakniknak: noted, will change

cores/grad3/hdl/jtgrad3_game.v
niknakniknak: id have to wire through sub too.  i'd prefer to leave this for now if its not a big deal.

cores/grad3/cfg/mame2mra.toml
niknakniknak: It should default to off, shouldn't it? The comment in dipsw.go says these bits are always 1, so bit 20 comes up as 1 and selects the second ids entry which is why I listed On first. The generated MRAs come out default="ff,5a,ff".
It does show as off on test builds on mister.

PR #1576, opened 2026-09-04
feat(): CRT analog resize for Mister

asturur:
This PR adapt a slimmed version of the crt resize from @Rmonic79 for jtframe.
The ability to stretch the analog signal without corrupting the image as we do today is very useful and a lot of crt users are growing accustomed to it.

I think is a good feature, so i modified that implementation for jtframe:
- removed the usage of bram in favor of a smaller buffer and a smaller analog stretching size
- calculated the stretching side on the base of JTFRAME macros so that all cores should get the correct buffer size and more or less the same amount of stretch.

the stretch goes from around -10 to 10% of size, it does not aim at reparing broken TVs, but just a bit to fit in or fill in a normal cosumer tv when 256px games tend to be extremely square or 384px games at less than 60hz go out of borders.

I kept it off for now for 31khz crt with the assumption that those monitors usually have controls for size on their own, and also because i don't have one for testing.

Vertical resizing is out of scope for now, is too much complicated and resource hungry

### Important

This replace the old resizer entirely, and leave and escape macro to use the old one.
The old one has no reason to be the default if this works.

### How the stretching works

For DIV=8 (standard 48 MHz / 6 MHz pixel clock), every pixel normally takes 8 master clocks. The scale controls how many pixels in each group of 8 get shifted to the adjacent divisor (7 or 9 clocks):

| scale | per group of 8 pixels | total clocks | % change |
|------:|:----------------------|-------------:|---------:|
|    -8 | 8 at 7 clocks        |           56 |   -12.5% |
|    -7 | 7 at 7, 1 at 8       |           57 |   -10.9% |
|    -6 | 6 at 7, 2 at 8       |           58 |    -9.4% |
|    -5 | 5 at 7, 3 at 8       |           59 |    -7.8% |
|    -4 | 4 at 7, 4 at 8       |           60 |   -6.25% |
|    -3 | 3 at 7, 5 at 8       |           61 |    -4.7% |
|    -2 | 2 at 7, 6 at 8       |           62 |    -3.1% |
|    -1 | 1 at 7, 7 at 8       |           63 |    -1.6% |
|     0 | 8 at 8 (bypass)      |           64 |       0% |
|    +1 | 7 at 8, 1 at 9       |           65 |    +1.6% |
|    +2 | 6 at 8, 2 at 9       |           66 |    +3.1% |
|    +3 | 5 at 8, 3 at 9       |           67 |    +4.7% |
|    +4 | 4 at 8, 4 at 9       |           68 |   +6.25% |
|    +5 | 3 at 8, 5 at 9       |           69 |    +7.8% |
|    +6 | 2 at 8, 6 at 9       |           70 |    +9.4% |
|    +7 | 1 at 8, 7 at 9       |           71 |   +10.9% |

The scale is literally how many of the 8 pixels shift to the adjacent divisor. At scale=-8, all 8 pixels move from 8→7 clocks (a full divisor shift). At scale=+7, 7 out of 8 pixels move from 8→9. The intermediate values blend linearly — the shifted pixels are distributed evenly across the line by the accumulator, never two adjacent.

It would be great if i could make the scale asymmetric and go from 8/8 to 8/8 pixel shifted, that can be done by changing the meaning of the 0 to 1 and remove the 0 from the scale ( since we have on/off ).

But for now let's evaluate the mechanism and the results and then we decide how to tweak it.


modules/jtframe/hdl/video/jtframe_hretime.v
```diff
@@ -0,0 +1,175 @@
...
+);
+
+localparam SL  = $clog2(STEP),                  // division by STEP is a shift
+           AW  = $clog2(DEPTH),
+           MW  = $clog2(DIV*(STEP+8))+1,        // rate accumulator width
+           PW  = SL+HW,                         // nactive * |scale|
+           DLD = 1<<$clog2((DEPTH*DIV)/2+1),    // sync delay line depth
+           DLW = $clog2(DLD);
+
+(* ramstyle = "MLAB, no_rw_check" *) reg [DW-1:0] mem[0:DEPTH-1];
```
asturur: here i decided to not use BRAM, differently from original implementation, because i hope those things at cost of ALM can be enabled on smaller fpga too. ( mist, sidi and sidi128 do not have the shifter either )

modules/jtframe/hdl/video/jtframe_hretime.v
```diff
@@ -0,0 +1,175 @@
...
+    { hs_dly, vs_dly} <= sync_mem[sync_wp-dly];
+end
+
+// Bypass path, kept aligned with the input pixel rate
+always @(posedge clk) if( ce_in ) begin
+    din_l <= din;
+    de_l  <= de_in;
+end
+
+endmodule
```
asturur, line 175: here there is lot to review!

modules/jtframe/hdl/video/jtframe_resync.v
```diff
@@ -9,8 +9,8 @@ module jtframe_resync(
     input         vs_in,
     input         LVBL,
     input         LHBL,
-    input  [3:0]  hoffset,
-    input  [3:0]  voffset,
+    input  [4:0]  hoffset,
+    input  [4:0]  voffset,
```
asturur, line 13: this 5 bit shift was a test, but once you scale i m not sure the 4 bits are good enough, we can leave this out of the final pr, but for the scope of testing the resize, having temporarly more space here is good.

modules/jtframe/target/mist/hdl/jtframe_mist_video.exp.v
```diff
@@ -0,0 +1,371 @@
...
+        video_hs <= 1'b1;
+        video_vs <= CSync_out;
+    end
+    if( cvideo_en ) begin
+        video_hs <= csync_c;
+        video_vs <= 1'b1;
+    end
+end
+`endif
+endmodule
```
asturur, line 371: this is an experimental copy of the framework with the function enabled, but i can't test it at all, i left here for discussion.

modules/jtframe/hdl/video/jtframe_hretime.v
jotego: For regular FPGA's using BRAM will result in less fitter effort and better signal timing.

I think we can bypass this module entirely for MiST and SiDi

asturur: Can you count how many alm are? i m ok moving it to bram, but i want still to put something on mist/sidi if possible.

asturur: i switched to BRAM. can you check this  is the way to signal quartus to use bram? (m9k or m10k)

modules/jtframe/target/mist/hdl/jtframe_mist_video.exp.v
asturur: deleted for now, i ll see what i can do for mist/sidi.

PR #1579, opened 2026-09-06
thundr: bring up Baraduke / Alien Sector

### LINKED ISSUE #1578: Baraduke / Alien Sector: wrong tilemap layer priority, and sound lock-up on speech
`aliensec` is currently in `parse.debug` in `cores/thundr/cfg/mame2mra.toml`, so
this is about the in-progress Baraduke support rather than a released game. Two
faults, both reproducible in simulation and confirmed on MiSTer hardware.

Reference throughout is MAME 0.289. Repro is a Verilator run of `cores/thundr`
with `aliensec`; the in-game faults need a cabinet script of 400 idle frames,
4 `coin`, 20 idle, 4 `1p`, then play.

### 1. A tilemap layer is drawn in front of the sprites

**Symptom:** during play a dark layer covers the scene and the player/enemy
sprites, which should be on top of it.

**Evidence:** same ROM, same inputs, same frame, compared against a MAME
snapshot of the same moment — 33.16% of pixels match (43119 differ). The
tilemap, HUD and shield bar are correct; the error is the layer ordering.

**Cause:** `jtcus42` applies the namcos86 rule, comparing the two layers'
priority fields (`priob > prioa`). baraduke.cpp does not do that — it selects
the background layer with

```c
int const back = (((m_tilegen->xscroll_r(0) & 0x0e00) >> 9) == 6) ? 1 : 0;
m_tilegen->draw(screen, bitmap, cliprect, back,     TILEMAP_DRAW_OPAQUE, 1);
m_tilegen->draw(screen, bitmap, cliprect, back ^ 1, 0,                   2);
```

so only layer A's priority field matters, and only whether it equals 6. The
generic comparison picks the wrong layer to draw in front.

Sprite priority differs on this board too: it is a single bit in `source[10]`
(`priority = source[10] & 0x01`), not the 3-bit field in `source[14]` that
namcos86 uses — that field reads back as zero here, so every sprite ends up at
priority 0. The associated rule is
`pri_mask = priority ? 0 : GFX_PMASK_2`, i.e. a priority-0 sprite hides behind
the front tilemap but not the back one.

### 2. Music stops permanently when the speech sample plays

**Symptom:** roughly 7 seconds into a new game, as the orange creature appears
and says "I'm your friend", the music stops and is replaced by a fixed tone that
persists until the core is reset. **Cabinet inputs keep working normally.**

**Evidence:** audio RMS per second from the simulation —

```
[0,0,0,0,0,0,0, 404,478,480,550,597,536,480, 214, 163,162,163,161,157,158, ...]
                └── game running ──────────┘  └── flat from t=14s onwards ──
```

MAME over the same window varies between 1800 and 5400 continuously. The stuck
output is not DC: it still has ~980 distinct sample values and ~3430
zero-crossings/sec, so the CUS30 is still generating a waveform — what has
stopped is anything updating it.

**Cause:** the MCU spins forever at `$84F2` in the speech player:

```
84f2: lda  $1105      ; CUS30 channel-0 voice byte 5
84f5: anda #$1F
84f7: lsra
84f8: cmpa $132A      ; compare against the value latched at routine entry
84fb: beq  $84F2      ; unchanged -> spin
```

The CUS30 voice registers occupy offsets `0x100-0x13F`, eight bytes per channel.
Bytes 0-4 are lvol/wsel/freq/rvol; **byte 5 reads back that channel's current
waveform position**. `jtcus30` returns plain RAM there, so the value never
changes.

Baraduke uses this as a timing reference: it programs channel 0 with a real
frequency but zero volume — MAME shows the writes `a=00 d=00`, `a=01 d=80`,
`a=02 d=95`, `a=03 d=2A`, `a=04 d=00` — purely so the position counter advances,
then polls byte 5 waiting for it to tick. Reading `$1105` in MAME returns 32
distinct values (0x00-0x1F) roughly uniformly; on the core it is static.

Controls survive because the enclosing output-compare ISR executes `CLI` at
`$F3A8` before the speech work, so vblank keeps being serviced while the sound
routine is wedged.

Note `jtcus30.sv` lives in `cores/shouse` and is shared with the shouse core, so
this affects any Namco game that uses the voice counter as a timing reference,
not only Baraduke.

### Related

The third bit plane of the scroll layer was also wrong for both sets — the
existing `# bad third bit plane of scroll layer` comment above the `debug` entry.
`bd?_6.4m` holds plane 2 for both tile layers, first half for layer A and second
half for layer B, with only the high nibble of each byte used; streaming the
whole ROM once per layer leaves plane 2 wrong in 5991 of 16384 tile rows.

memmerson:
Fixes #1578

Brings up Baraduke / Alien Sector on JTTHUNDR and fixes the two faults reported
there. The game runs under the `aliensec` parent set, which shares baraduke.cpp
and its `machine_config` with Metro-Cross, so most of the existing Metro-Cross
path is reused — the changes below are the places where that board genuinely
differs from namcos86.

### Layer priority — `jtcus42.v`

`jtcus42` applied the namcos86 rule, comparing the two layers' priority fields
(`priob > prioa`). baraduke.cpp does not do that; it selects the background layer
with

```c
int const back = (((m_tilegen->xscroll_r(0) & 0x0e00) >> 9) == 6) ? 1 : 0;
m_tilegen->draw(screen, bitmap, cliprect, back,     TILEMAP_DRAW_OPAQUE, 1);
m_tilegen->draw(screen, bitmap, cliprect, back ^ 1, 0,                   2);
```

so only layer A's priority field matters, and only whether it equals 6. The
generic comparison picked the wrong layer to draw in front, covering the sprites.
Selected by the existing `metrocrs` header bit. `frontop` is exported so the
colour mixer knows where the front layer is opaque.

### Sprite priority — `jtthundr_objscan.v`, `jtmetrox_colmix.v`

On this board sprite priority is a single bit in `source[10]`, not the 3-bit
field in `source[14]` that namcos86 uses — that field reads back as zero here, so
every sprite ended up at priority 0. The associated rule is
`pri_mask = priority ? 0 : GFX_PMASK_2`, i.e. a priority-0 sprite hides behind the
front tilemap but not the back one. The priority bit doubles as `code[8]`, which
is harmless because the sprite region is mirrored.

### CUS30 voice counter read-back — `jtcus30.sv`

The CUS30 voice registers occupy offsets `0x100-0x13F`, eight bytes per channel.
Bytes 0-4 are lvol/wsel/freq/rvol; byte 5 reads back that channel's current
waveform position. `jtcus30` returned plain RAM there, so the value never
changed.

Baraduke uses this as a timing reference for speech: it programs channel 0 with a
real frequency but zero volume purely so the position counter advances, then
polls byte 5 waiting for it to tick, spinning forever at `$84F2` without it. The
enclosing output-compare ISR executes `CLI` first, which is why the cabinet
inputs kept working while the music was stuck.

Note the packed config arrays are indexed in reverse: channel 0 is element 7.

`jtcus30.sv` lives in `cores/shouse` and is shared with the shouse core, so this
affects any Namco game using the voice counter as a timing reference — see the
regression note below.

### Third bit plane — `mame2mra.toml`

`bd?_6.4m` holds plane 2 for both tile layers: its first half feeds layer A
(gfx0), its second half layer B (gfx1), and only the high nibble of each byte is
used. Each half is now streamed twice, tiles 0-511 taking the high nibble and
512-1023 the low one, selected by `code[9]` — see `init_baraduke()`. Previously
the whole ROM was streamed once per layer, leaving plane 2 wrong in 5991 of 16384
tile rows. This is the `# bad third bit plane of scroll layer` note above the
existing `debug` entry.

Also names the single button "Blaster" for `aliensec`.

### DIP switches — `jtmetrox_cab.v`

Corrected to the MAME port layout, where `dipsw[7:0]` is DSWA, `[15:8]` DSWB and
`[19:16]` EDGE. The previous code packed the fields contiguously, so every OSD
option mapped to the wrong physical switch. Invisible at factory defaults, since
those are all ones — which is presumably why it went unnoticed.

### Minor

- `jtmetrox_text.v` — split a declaration mixing an initialised and an
  uninitialised net, which iverilog rejects. No functional change.
- `jtthundr_obj.v`, `jtthundr_video.v` — plumb `metrocrs` and `frontop` through.
- `ver/aliensec` — `trace.mame` moved from `ver/baraduke` to the parent setname,
  `main.mame` added for the 6809 in the CSV form `jtutil trace` expects, and the
  year in the MAME filter corrected to 1985.

## Verification

Against MAME 0.289, comparing palette indices rather than RGB (jtframe replicates
nibbles where MAME models the resistor ladder, so the values differ by design):

- **Baraduke** — an attract frame matches the reference **64512/64512 pixels**
  (100.00%). With the old tile packing the same frame was 59.95%.
- **Metro-Cross** — matches at 100.00% at four separate points in its attract
  loop, and is **byte-identical over 700 frames** with and without these changes,
  so the priority rework is a no-op for it.
- **Sound** — audio continues past the speech event and tracks the reference
  envelope, where it previously pinned to a constant level from t=14s onwards.
  The first 14s of audio are unchanged.

Because `jtcus30.sv` is shared, it was regressed against shouse as well:
**pacmania, splatter, galaga88 and dspirit are byte-identical in both video and
audio**, with and without the change, over 700 frames each. All four boot,
render and produce sound in those runs, so the comparison isn't vacuous. That is
4 of the 20 setnames in `shouse/cfg/reg.yaml`, not the full set.

Both fixes are confirmed on MiSTer hardware.

## Not addressed

A small cosmetic artifact remains on the Baraduke title screen: 120 pixels at
x 251-258, purely additive (nothing is missing). Attributed to the CUS42 tilemap
by re-running with `JTFRAME_SIM_GFXEN` to disable the sprite and text layers in
turn — the artifact survives both. Likely an off-by-one at the tilemap's
horizontal wrap, which would be invisible wherever the map is uniform, consistent
with gameplay frames matching at 100%. Left alone rather than guessing at
`jtframe_scroll`, which is shared widely.

`aliensec` is left in `parse.debug`; removing it is a release decision.

## Disclosure

This was investigated and written with the help of Claude (Anthropic's coding
agent). I don't have Verilog experience myself, so I can't personally vouch for
the HDL — please review it accordingly, and I'm happy to change or drop any part.
The rationale for each change is given above against the MAME driver behaviour it
derives from, so it should be checkable without taking my word for it. The
testing above I ran and checked myself.


asturur:
i would like to test the blast-radius action on this pr with the test executions, seems a good target.

memmerson:
> i would like to test the blast-radius action on this pr with the test executions, seems a good target.

Go right ahead, happy for this to be a guinea pig

asturur:
ok what i wanted to test was the action:

Cores possibly affected: **paclan**, **shouse**, **thundr**.

jotego:
Thank you! Can you check past frame 700? That's too few. Please extend the check up to 2000 frames at least.

memmerson:
> Thank you! Can you check past frame 700? That's too few. Please extend the check up to 2000 frames at least.

Done — re-ran the whole set at 2000 frames.

**Method:** two worktrees, PR head `f221881a4` against its parent `77a10e72d`, same ROMs, `jtsim -verilator -load -video 2000` on each. Dumped frames and audio compared byte-for-byte. One simulation at a time — an earlier batch ran several in parallel and the frame counts weren't reproducible, solo runs are.

| set | core | frames compared | differing | audio |
|---|---|---|---|---|
| metrocrs | thundr | 1460 | 0 | identical |
| pacmania | shouse | 1767 | 0 | identical |
| splatter | shouse | 1408 | 0 | identical |
| galaga88 | shouse | 1713 | 0 | identical |
| dspirit | shouse | 1240 | 0 | identical |
| pacland | paclan | 1044 | **9** | **differs** |

"Frames compared" is the number of distinct frames dumped across the 2000-frame run — `jtsim` skips repeats, so an attract screen holding still produces one frame, not two hundred.

**pacland is the one real difference,** and I think it's a correction rather than a regression. Frames 67, 68, 69, 164, 933, 939, 940, 1805 and 1807 differ, and the audio differs throughout. All of it comes from the `jtcus30.sv` hunk — voice register byte 5 reading back the live waveform counter instead of the last value written to sound RAM.

Evidence it's the right way round: I put read and write taps on `namco_cus30_device` in MAME and ran pacland over the same stretch. There were 16 read-after-write accesses to a voice byte 5, and in **0** of them did the read return the byte that had just been written — MAME returned the counter every time. So the pre-patch core was wrong for pacland as well; it just never mattered visibly because nothing in pacland polls that register in a loop.

Baraduke does: the MCU's speech routine spins at `$84F2` polling `$1105` (voice 4, byte 5) and never leaves the loop while the value is static. That's the sound lock-up in #1578.

I have not checked pacland on hardware — only in simulation and against MAME. If you'd rather not carry an audio change for pacland in this PR, I'm happy to split the `jtcus30.sv` hunk into its own PR so the thundr work can go in independently.

On the blast-radius output — paclan, shouse and thundr is exactly right. `cores/shouse/hdl/jtcus30.sv` is the only shared file the PR touches.

cores/shouse/hdl/jtcus30.sv
```diff
@@ -174,6 +177,15 @@ jtframe_dual_ram u_wave(
...
 );
 
+// Voice byte 5 reads back that channel's live waveform position. Baraduke's
+// speech routine polls it as a timing reference and hangs without it.
+// NB the packed arrays are indexed in reverse: channel 0 is element 7.
+always @(posedge clk) begin
+    cntrd_l <= xaddr[9:6]==4'b0100 && xaddr[2:0]==3'd5;
+    cntch_l <= 3'd7 - xaddr[5:3];
+end
+assign xdin = cntrd_l ? { 3'd0, cnt[cntch_l][CW-1:A0] } : ram_q0;
```
jotego: Register xdin instead of the combinational assignment

cores/shouse/hdl/jtcus30.sv
```diff
@@ -48,6 +48,9 @@ wire        xwe   = xcs & ~(bsel ? brnw : srnw);
 wire        mmr_cs= (bsel ? bcs  : scs) && xaddr[9:6]==4'b0100;
 wire        mmr_wn= ~mmr_cs | (bsel ? brnw : srnw);
 
+wire [ 7:0] ram_q0;
+reg         cntrd_l;
```
jotego: Rename cntrd_l to cntpos_cs

cores/shouse/hdl/jtcus30.sv
```diff
@@ -48,6 +48,9 @@ wire        xwe   = xcs & ~(bsel ? brnw : srnw);
...
+reg  [ 2:0] cntch_l;
```
jotego: Rename to chsel or just ch

cores/shouse/hdl/jtcus30.sv
```diff
@@ -48,6 +48,9 @@ wire        xwe   = xcs & ~(bsel ? brnw : srnw);
 wire        mmr_cs= (bsel ? bcs  : scs) && xaddr[9:6]==4'b0100;
 wire        mmr_wn= ~mmr_cs | (bsel ? brnw : srnw);
 
+wire [ 7:0] ram_q0;
```
jotego: Rename to xwave

cores/shouse/hdl/jtcus30.sv
```diff
@@ -174,6 +177,15 @@ jtframe_dual_ram u_wave(
     .q1     ( wdata8     )
 );
 
+// Voice byte 5 reads back that channel's live waveform position. Baraduke's
```
jotego: Delete the comments

cores/thundr/cfg/mame2mra.toml
```diff
@@ -116,20 +117,44 @@ regions = [
 		{name="mc1-6.4n", crc="9686dc3c",map="2100"},
 	], mirror=true },
 	# aliensec tiles
+	# The third bit plane lives in bd2_6.4m: its first half feeds tile layer A
```
jotego: Remove the comments

cores/thundr/cfg/mame2mra.toml
```diff
@@ -116,20 +117,44 @@ regions = [
...
+	# nibble of each byte is used, so each half is streamed twice: tiles 0-511
+	# take the high nibble, tiles 512-1023 the low one, selected by code[9] in
+	# jtthundr_scroll. See init_baraduke() in baraduke.cpp.
 	{ name="tiles_0", start="JTFRAME_BA2_START", setname="aliensec", parts=[
-		{name="bd2_8.4p", crc="432bd7d9",map="0021"},
-		{name="bd2_6.4m", crc="f4c1df60",map="2100"},
-
-		{name="bd1_7.4n", crc="0d7ebec9",map="0021"},
-		{name="bd2_6.4m", crc="f4c1df60",map="2100"},
+		# layer A - tiles 0-511
```
jotego: Remove the comments or at least leave just two: layer A and layer B

cores/thundr/cfg/mame2mra.toml
```diff
@@ -116,20 +117,44 @@ regions = [
...
+		{name="bd2_6.4m", crc="f4c1df60",map="2100", offset=0x2000, length=0x2000},
+		# layer B - tiles 512-1023
+		{name="bd1_7.4n", crc="0d7ebec9",map="0021", offset=0x2000, length=0x2000},
+		{name="bd2_6.4m", crc="f4c1df60",map="2100", offset=0x2000, length=0x2000},
 	], mirror=true },
 	# baraduke tiles
+	# The third bit plane lives in bd1_6.4m: its first half feeds tile layer A
+	# (MAME gfx0), its second half feeds layer B (gfx1). Only the high
+	# nibble of each byte is used, so each half is streamed twice: tiles 0-511
+	# take the high nibble, tiles 512-1023 the low one, selected by code[9] in
```
jotego: Maybe this can be done without explicit file names

cores/thundr/hdl/jtcus42.v
```diff
@@ -7,6 +7,7 @@ module jtcus42(
     input               rst,
     input               clk, pxl_cen, dec_en, plane3inv,
     input               flip, scrhflip, hs,
+    input               metrocrs,       // baraduke.cpp board mixing rules
```
jotego: Remove comment

cores/thundr/hdl/jtcus42.v
```diff
@@ -47,11 +49,16 @@ reg  [ 8:0] effxa, effxb;
...
 wire [ 7:0] scrya, scryb, adec_data, bdec_data;
 wire [ 4:0] adec_addr, bdec_addr;
-wire        scra_op, scrb_op, selb;
+wire        scra_op, scrb_op, selb, backb;
 
 assign scra_op = scra_pxl[2:0]!=ALPHA;
 assign scrb_op = scrb_pxl[2:0]!=ALPHA;
-assign selb    = scrb_op && (!scra_op || priob > prioa);
+// baraduke.cpp draws one layer opaque and the other one over it. The background
+// is layer B only when layer A's priority field reads 6, see screen_update()
```
jotego: Remove comment

cores/thundr/hdl/jtmetrox_cab.v
```diff
@@ -23,11 +23,12 @@ reg [2:0] sel;
 
 always @(posedge clk) begin
     if(&p1_dout[7:5]) sel=p1_dout[2:0];
+    // dipsw follows the MAME port layout: [7:0]=DSWA, [15:8]=DSWB, [19:16]=EDGE
     case(sel)
-        0: cab <= {UNUSED,dipsw[4:0]};
-        1: cab <= {UNUSED,dipsw[7:5],dipsw[9:8]};
-        2: cab <= {UNUSED,dipsw[14:10]};
-        3: cab <= {UNUSED,dipsw[15],dipsw[19:16]};
+        0: cab <= {UNUSED,dipsw[ 7:3]};              // (DSWA & 0xf8)>>3
```
jotego, line 28: Why is this changed?

cores/thundr/hdl/jtmetrox_colmix.v
```diff
@@ -23,14 +24,17 @@ localparam [2:0] ALPHA=7;
 localparam [3:0] OALPHA=15;
 localparam [1:0] OBJ=2'b01;
 
-wire       txt_op, obj_op;
+wire       txt_op, obj_op, obj_win;
 
 assign txt_op = txt_pxl[1:0]!= ALPHA[1:0] && gfx_en[0];
 assign obj_op = obj_pxl[3:0]!=OALPHA      && gfx_en[3];
+// baraduke.cpp draws the sprites with pri_mask = priority ? 0 : GFX_PMASK_2, so
+// a sprite with priority 0 hides behind the front tilemap but not the back one
```
jotego: Remove comment

cores/thundr/hdl/jtmetrox_text.v
```diff
@@ -28,7 +28,9 @@ wire [10:1] vram_prea;
 wire [ 8:0] code, hadj, vadj;
 wire [ 6:0] pal;
 wire [ 2:0] code_msb;
-wire        blankn = hdump>=9'o60 && hdump<9'o550, border;
+wire        border;
+// split declaration: iverilog rejects `wire a = expr, b;`
```
jotego: Remove comment

cores/thundr/hdl/jtthundr_objscan.v
```diff
@@ -80,11 +80,18 @@ end
 
 always @(posedge clk) if(cen) begin
     case(st)
-        0: { code[7:0], hsize, hflip, hos, code[10:8] } <= ram_dout;
+        0: begin
+            { code[7:0], hsize, hflip, hos, code[10:8] } <= ram_dout;
+            // baraduke.cpp: priority = source[10] & 1. That bit doubles as
```
jotego: Remove comment

cores/thundr/hdl/jtthundr_objscan.v
```diff
@@ -80,11 +80,18 @@ end
...
+            // code[8] but the sprite ROM is mirrored, so code[10:7] are moot
+            bara_prio <= ram_dout[0];
+        end
         1: { hpos[7:0], pal, hpos[8] } <= ram_dout;
         2: begin
             wide  <= hsize==HLARGE;
-            { prio, vos, vsize, vflip } <= ram_dout[7:0];
+            { vos, vsize, vflip } <= ram_dout[4:0];
+            // namcos86 keeps a 3-bit priority in source[14], baraduke.cpp does not
```
jotego: Remove comment

jotego (changes requested):

asturur:
@memmerson i ll be the bad one saying this, if your comments are prepared with the AI, they are extremely heavy to read and parse. Read them, condense them as a human and write them with your own language, at least we are having a conversation.

Would have you write:


> Done — re-ran the whole set at 2000 frames.
> 
> **Method:** two worktrees, PR head `f221881a4` against its parent `77a10e72d`, same ROMs, `jtsim -verilator -load -video 2000` on each. Dumped frames and audio compared byte-for-byte. One simulation at a time — an earlier batch ran several in parallel and the frame counts weren't reproducible, solo runs are.
> 
> | set | core | frames compared | differing | audio |
> |---|---|---|---|---|
> | metrocrs | thundr | 1460 | 0 | identical |
> | pacmania | shouse | 1767 | 0 | identical |
> | splatter | shouse | 1408 | 0 | identical |
> | galaga88 | shouse | 1713 | 0 | identical |
> | dspirit | shouse | 1240 | 0 | identical |
> | pacland | paclan | 1044 | **9** | **differs** |
> 
> "Frames compared" is the number of distinct frames dumped across the 2000-frame run — `jtsim` skips repeats, so an attract screen holding still produces one frame, not two hundred.
> 
> **pacland is the one real difference,** and I think it's a correction rather than a regression. Frames 67, 68, 69, 164, 933, 939, 940, 1805 and 1807 differ, and the audio differs throughout. All of it comes from the `jtcus30.sv` hunk — voice register byte 5 reading back the live waveform counter instead of the last value written to sound RAM.

Or something like:

> Done, i did regression testing with the commit before and everything is identical with jtsim comparison apart pacland that has 9 differences. But those 9 differences seems legit to me because ....

The issue with AI comments is that if someone find them not useful 1-2-3 times and then starts to recognize the english pattern in the first 2-3 lines, this person will start to skip them, run over the lines see if there is some other information later on, no one is going to read them.

Then if i m wrong, i m very sorry i talked too soon, but those seems ai comments to me.

jotego:
I personally cannot stand them.

memmerson:
Comments are AI, and I accept and agree with what you are saying. In my line of work, devops, overly verbose comments would irk me also. For context, I am not an FPGA developer I certainly couldn't do this work unassisted, I have used Claude to get a couple of my favourite games (Baraduke and Wardner) working on Mister, once I (with Claude) got them working I have posted the commits here for the verification of real experts (you guys) so we can hopefully, eventually share with the community.
The comments are absolutely AI summaries and clearly more verbose than is required, you can blame my lack of FPGA experience for that and I am grateful that you guys have taken the time to review this work.
I will work through and make the requested changes and will also ensure future comments are more succinct!
:-)

cores/thundr/hdl/jtmetrox_cab.v
memmerson: (Claude's response) The old code read the wrong DIP bits. baraduke.cpp builds the four nibbles from DSWA 7-3, DSWA 2-0 + DSWB 7-6, DSWB 5-1, and DSWB 0 + EDGE. The MRA puts DSWA at bits 0-7 and DSWB at 8-15, so Bonus Life (bits 14-15) was read from bits 8-9. Metro-Cross is affected too.

cores/thundr/cfg/mame2mra.toml
memmerson: (Claude's response) I don't think it's possible with the current options. parts needs the file name, and sequence with width=32 puts one byte per file in each lane. These tiles need two bytes from the same file per word, so it would mean changing the scroll ROM layout that the other namcos86 games share.

cores/shouse/hdl/jtcus30.sv
memmerson: (Claude's comment) I tried this, but it adds a clock of latency to all CUS30 reads and changes pacland (from frame 67), metrocrs and aliensec in simulation. dspirit is unaffected. The select signals are already registered and the RAM output is synchronous, so I left the final mux combinational. Happy to do it another way if you prefer.

jotego:
Thank you! I know that steering the AI requires time and effort too. I appreciate it that you took the time to share the edits. 🙂

PR #1592, opened 2026-09-09
fix(hq2x): add line buffer to NOHQ2X scandoubler bypass

### LINKED ISSUE #1582: system 16 scan doubler problem
In the latest versions of the System16 and System16B cores, it looks like something has changed in the video output implementation. When Double Scan is enabled to generate a signal compatible with 31 kHz CRT monitors by doubling the vertical resolution, the image appears doubled both over HDMI and through the analog output.

Initially, you mentioned that the problem could be related to changes in the MiSTer main/master branch. However, I have been testing versions of the JTS16 and JTS16B cores from about two months ago, and Double Scan works perfectly with those versions. This suggests that something has been modified in the cores themselves that is causing Double Scan to no longer work correctly.

Something similar happens with the CPS3 core. Double Scan works correctly in the CPS2 core, but it has never worked properly in CPS3 and has always produced a doubled image. i think is related to some people having scaler issues

Therefore, it seems that there may have been a recent change in the video output implementation of these cores that is causing this issue.

asturur:
The JTFRAME_NOHQ2X ifdef (68e506d7d) replaced the full Hq2x module with a 2-register passthrough that had no line buffer.

The scandoubler timing still doubled the sync/blank signals, but pixel data was sampled live from the input — the second output line showed the start of the NEXT input line instead of replaying the current one, producing out-of-phase doubled lines on screen.

I decided to divide the problem.
Instead of making hq2x a deactivate module with passthrough i amde the scandoubler not include instantiate the module in case of JTFRAME_NOHQ2X and then i made a small libe buffer into the scandouble, always inside the scandoubler.

Adding a line buffer inside hq2x ( the first iteration ) was confusing.

affected cores: s16, contra, castle, cps3, mx5k, sf

Fixes jotego/jtcores#1582

### DIFF
--- a/modules/jtframe/target/mister/hdl/sys/hq2x.sv
+++ b/modules/jtframe/target/mister/hdl/sys/hq2x.sv
@@ -29,14 +29,6 @@ module Hq2x #(parameter LENGTH, parameter HALF_DEPTH)
 localparam AWIDTH = $clog2(LENGTH)-1;
 localparam DWIDTH = HALF_DEPTH ? 11 : 23;
 localparam DWIDTH1 = DWIDTH+1;
-`ifdef JTFRAME_NOHQ2X
-	reg [DWIDTH:0] pxlin_l, pre_out;
-	assign outpixel = pre_out;
-	always @(posedge clk) begin
-		if(ce_in ) pxlin_l <= inputpixel;
-		if(ce_out) pre_out <= pxlin_l;
-	end
-`else
 
 (* romstyle = "MLAB" *) reg [5:0] hqTable[256];
 initial begin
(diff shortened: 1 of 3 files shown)


asturur:
List of affected cores: all.

asturur:
@jotego i deliberately did some changes that weren't really necessary but seemed cleaner code to me.

When NOHQ2X is used, instead of disabling NHOQ2X module internally i avoid instantiating it.
The missing line buffer provided from NOHQ2X is now under macro in the scandoubler file with the extra lines of code the AI added and that i think are the minimum i can add.

This to me seems more clear on how it works.

asturur:
confirmed working on arcade 31khz  crt.

asturur:
Need to double check CPS2 STA failures, maybe what i added adds a longer combinational path than it should

jotego:
I think the purpose of the macro is to help with timing problems inside the HQ2X filter, particularly at 96MHz.
Instead of fixing them, it just bypasses the filter as it is a feature that a lot of people don't like and only a few people use.

asturur:
> I think the purpose of the macro is to help with timing problems inside the HQ2X filter, particularly at 96MHz. Instead of fixing them, it just bypasses the filter as it is a feature that a lot of people don't like and only a few people use.

Yes it is. and that didn't change with my pr, it is just approached slightly differently. I think hq2x make no sense in general but that is another topic.

PR #1596, opened 2026-09-10
feat(): Add Wardner core

memmerson:
**Title:** Wardner (Toaplan TP-009 / Taito B25) core, plus a TMS320C10 module

## What this adds

A core for Toaplan's TMS320C10 boards, starting with Wardner (1987). It runs on
a DE10-Nano — video, audio and controls — and is verified against MAME 0.289 at
every level.

Two commits:

- `doc: regenerate mame.xml from MAME 0.289` — the vendored XML carried no
  Wardner machine, so `jtframe mra` could not run for this core. Regenerated via
  `jtframe mra --reduce`; strictly additive, no machine removed, 18 added.
- `wardner: Toaplan TP-009 / Taito B25 arcade core` — the core, the DSP module
  and four test benches.

## The hardware, and why it is worth more than one game

TP-009 is the Flying Shark / Twin Cobra video and DSP hardware with a Z80 in
place of the 68000:

- Z80 @ 6 MHz (24 MHz XTAL / 4), banked ROM at 0x8000
- Z80 @ 3.5 MHz + YM3812, talking to the main CPU through 2 KB of shared RAM.
  No sound latch, no NMI — the sound CPU polls.
- TMS320C10 @ 14 MHz. **Not protection.** It halts the main Z80, takes the bus,
  and writes enemy fire, collisions and sprite placement straight into the Z80's
  work RAM, sprite RAM and palette RAM. Nothing runs without it.
- Toaplan SCU sprites, three 8x8 tilemaps, HD6845S, 446x286, 320x240, 54.878 Hz

Everything but the main CPU and the DSP's host-address decoder is shared with
Twin Cobra, Flying Shark, Sky Shark and Demon's World. Board-level modules are
named accordingly (`jttoaplan1_dsp`, `modules/jt32010`) so those become "swap
the CPU and the decoder" rather than new cores.

`modules/jt32010` is a TMS320C10 written from scratch — ~60 instruction forms,
two auxiliary registers, a four-level stack, a 16x16 multiplier — in the pattern
of `jt6295` / `jt7759`. It also unlocks Demon's World and the BSMT2000.

## Verification

Every figure below is printed by the bench that produces it.

| | |
|---|---|
| `jt32010` | 81 programs x 20000 instructions against a C model transcribed from MAME |
| DSP wrapper | 6/6 transaction cases; 9/9 deliberate mutations caught; the real Wardner program for 480000 instructions over 16 host-RAM seeds |
| main Z80 | boots the real ROM through POST and the DSP handshake |
| sound | 288 notes keyed on; envelopes track MAME at r = 0.95–0.98; measured output rate 48611 Hz = 3.5 MHz / 72 exactly |
| video | pixel-exact against a MAME-transcribed reference on 20 game snapshots and 76 random ones, both flip states, and against MAME's own captured frames |
| whole core | 54.8814 Hz under jtsim with the real SDRAM controller; every ROM region byte-exact |
| MRA | generated `.rom` byte-identical to the known-good image except two DSP words (see below) |
| synthesis | 5CSEBA6U23I7, 61% logic, 27% RAM blocks, timing closed |

### Reproducing it

```bash
source modules/jtframe/bin/setprj.sh
git submodule update --init                      # jtopl for the YM3812

modules/jt32010/ver/cpu/regress.sh               # ~2 min
cores/wardner/ver/dsp/run.sh                     # transaction conformance
cores/wardner/ver/dsp/run_real.sh  $ROMDIR       # the real DSP program
cores/wardner/ver/main/run.sh      $ROMDIR 3000  # boots the real ROM
cores/wardner/ver/video/fuzz.sh    24 1          # frame diff
jtcore wardner -mister --nodbg
```

`$ROMDIR` is a directory holding a Wardner set; the bootleg `wardnerb` works for
everything except the MRA. Nothing ROM-derived is committed.

## Questions for review

1. **Module naming.** Game-side is `jtwardner_*`, board-side `jttoaplan1_dsp`,
   CPU `jt32010`. Happy to rename any of it.
2. **The DSP dump the MRA carries.** Wardner's mask ROM has never been read;
   MAME substitutes the Flying Shark part and flags it `BAD_DUMP`. The bootleg
   sets carry Wardner's own program across eight nibble PROMs, which `mame2mra`
   cannot reassemble (no nibble-interleave transform), so the MRA can only build
   the parent's. The two differ in exactly two words — one bit each, the
   direct/indirect addressing flag at `$034` and `$36d` — and nothing testable
   reaches either address. Would you take a PR adding a nibble transform to the
   region options, or is shipping the parent's fine?
3. **`ver/` is force-added.** `.gitignore` carries `/cores/*/ver/**`, so the 36
   bench files here go against that policy. Happy to drop them if benches are
   meant to stay untracked.
4. **`[mist|sidi] JTFRAME_SKIP` is a prediction.** No Cyclone IV E package or
   Quartus 13 available here, so the EP3C25 fit is untested. Happy to drop the
   skip if it fits.
5. **Screen flip matches MAME, and MAME looks wrong.** `flipscreen_w` touches
   only the tilemaps, so a flipped Wardner shows mirrored tilemaps over
   unmirrored sprites, and the flipped view is displaced 79 px horizontally and
   213 vertically rather than mirrored. Reproduced faithfully so the diff stays
   meaningful, but I would not offer cocktail mode until someone checks a real
   board.
6. **The sprite-per-line limit is mine, not the board's** — roughly 70 per line
   from the line's clock budget. MAME has no limit; it never triggers in play.

Also missing: no `cfg/reg.yaml`.

## One thing worth flagging about the benches

The video reference renderer is a transcription of `twincobr_v.cpp` and
`toaplan_scu.cpp` made by the same hand as the RTL, and for a long time it
carried the same misreading of MAME's `gfx_layout` — that `planeoffset` is
listed from the most significant pen bit down. Both sides had every pen bit
reversed. Bit reversal maps 0 to 0, so pen 0 stayed transparent and every shape,
silhouette and priority decision was exactly right while every non-zero pen took
the wrong colour. The frame diff was comparing two copies of one mistake and
reported 0 of 76800 pixels on every snapshot while the picture was visibly
wrong. It only surfaced on real hardware.

The benches now check against MAME's own captured output as well as against each
other. `doc/plan.md` section 19 covers this and the four other bugs that reached
hardware — every one of them in the seam between the emulation and the framework
rather than in the emulation.

## A note on authorship

I'm not an FPGA or Verilog developer. The HDL, the benches and the analysis in
this PR were written by Claude (Anthropic's Claude Code) working from the MAME
0.289 sources, under my direction. I picked the target, ran the core on my own
DE10-Nano, and reported the faults that only showed up there — most of which
the simulation benches could not see.

Every figure above is printed by a bench you can run yourself, and
`doc/plan.md` marks each claim as either verified against a primary source or
inferred. Happy to rework anything that doesn't fit.


cores/wardner/cfg/mem.yaml
```diff
@@ -0,0 +1,65 @@
...
+    - freq: 6M
+      outputs:
+        - cen6
+    # the 14 MHz crystal, as a fractional chain: 48 * 7/24. The board derives
+    # the pixel clock and the audio/CRTC clock from it by the same halving.
+    - mul: 7
+      div: 24
+      outputs:
+        - cen14           # TMS320C10
+        - cen7            # pixel clock
```
asturur, line 17: the pixel clock is better when completely in sync with the real clock.
If you have a pixel-clock of 7Mhz verified, use the JTFRAME macro to use the pll_700 and the macro SHIFT_180 ( look at the cps3 core to see how is done ).

asturur: This will move your base clock away from 48Mhz, so then instead of mul and div you will have to use freq: 14M.

cores/wardner/cfg/mem.yaml
```diff
@@ -0,0 +1,65 @@
...
+          offset: FG_OFFSET
+    - buses:
+        - name: obj
+          addr_width: 18        # 65536 rows, 16x16 sprites are two per row
+          data_width: 32
+audio:
+  rsum: 10k
+  rc: { r: 1.3k, c: 10n }
+  channels:
+    - { name: fm, module: jtopl, rsum: 6.2k, rc: [{ r: 1rout, c: 1n }] }
```
asturur: usually we put audio under clocks, the sdram, then bram. move this up.

cores/wardner/cfg/mem.yaml
```diff
@@ -0,0 +1,65 @@
...
+bram:
+  # TMS320C10 mask ROM: 0x1000 bytes in the MAME region, big endian, of which
+  # the game uses 0x600 words. Loaded straight out of the download stream, so
+  # the DSP never competes for SDRAM - it fetches every machine cycle.
```
asturur: One line comments are ok, or more if they say something that is useful. this is not useful. i would remove it or remove the endianness, how much the game uses, how is loaded etc etc.

cores/wardner/cfg/msg
```diff
@@ -0,0 +1,16 @@
...
+          1987
+
+Z80 main and sound CPUs, a
+TMS320C10 co-processor, YM3812
+and the Toaplan sprite unit.
+
+The DSP is not protection: it
+runs enemy fire, collisions and
+sprite placement, writing them
+into the main CPU's own RAM.
```
asturur: Here usually we write about the core and how it has been made, like 'Built from mame sources'
Check a standard core of the latest done to get an idea.

cores/wardner/doc/plan.md
asturur, line 1: This file is not needed in the repo.

cores/wardner/hdl/jtwardner_game.v
```diff
@@ -0,0 +1,279 @@
...
+// that to read left, right, up, down from bit 3 down, joy_in[3] must be up and
+// joy_in[0] right. So jtframe gives up, down, left, right in bits 3 to 0 while
+// the board wants them in bits 0 to 3, and the direction nibble is reversed.
+// Buttons are in the same place in both.
+//
+//   j[3] up    -> bit 0        j[1] left  -> bit 2
+//   j[2] down  -> bit 1        j[0] right -> bit 3
+function [7:0] toaplan_joy(input [5:0] j);
+    toaplan_joy = { 2'b00, ~j[5], ~j[4], ~j[0], ~j[1], ~j[2], ~j[3] };
+endfunction
```
asturur: i have been told more than once to do not do functions for 1 liners.
Check fi this can be done as you did cab_sys

cores/wardner/hdl/jtwardner_game.v
```diff
@@ -0,0 +1,279 @@
...
+
+// The foreground bank bit selects tiles 4096-8191. Wardner's foreground ROM
+// holds exactly 4096 tiles, and MAME wraps the code with code % total, so
+// dropping the bit is what the reference does. Twin Cobra has twice the ROM
+// and would use it, which is why the video module keeps the bit.
+assign fg_addr = fg_full_addr[14:0];
+assign dsprom_addr = dsp_rom_addr[10:0];
```
asturur: Not sure if those 2 assign make sense. I m not sure, verify with AI fi dsprom_addr is unused, and if fg_addr can be just inlined.

cores/wardner/hdl/jtwardner_main.v
```diff
@@ -0,0 +1,348 @@
...
+jtframe_dual_ram #(.AW(11),.DW(8)) u_obj_hi(
+    .clk0(clk),.data0(sh_dhi),.addr0(sh_addr),.we0(obj_hi_we),.q0(obj_q[15:8]),
+    .clk1(clk),.data1(8'd0  ),.addr1(obj_vaddr),.we1(1'b0    ),.q1(obj_vq[15:8]) );
+
+jtframe_dual_ram #(.AW(11),.DW(8)) u_pal_lo(
+    .clk0(clk),.data0(sh_dlo),.addr0(sh_addr),.we0(pal_lo_we),.q0(pal_q[ 7:0]),
+    .clk1(clk),.data1(8'd0  ),.addr1(pal_vaddr),.we1(1'b0    ),.q1(pal_vq[ 7:0]) );
+jtframe_dual_ram #(.AW(11),.DW(8)) u_pal_hi(
+    .clk0(clk),.data0(sh_dhi),.addr0(sh_addr),.we0(pal_hi_we),.q0(pal_q[15:8]),
+    .clk1(clk),.data1(8'd0  ),.addr1(pal_vaddr),.we1(1'b0    ),.q1(pal_vq[15:8]) );
```
asturur: Ask AI if this stuff is generic RAM for the z80 it goes defined in mem.yaml

asturur: Verify for all the jtframe_dual_ram in this file. scratch ram internal to dsps are ok inside the module. board level rams go into mem.yaml

cores/wardner/hdl/jtwardner_main.v
```diff
@@ -0,0 +1,348 @@
...
+always @(posedge clk, posedge rst) begin
+    if( rst ) begin
+        irq_n  <= 1'b1;
+        lvbl_l <= 1'b1;
+    end else begin
+        lvbl_l <= LVBL;
+        if( !int_en ) irq_n <= 1'b1;
+        else if( lvbl_l && !LVBL ) irq_n <= 1'b0;   // blanking has just begun
+    end
+end
```
asturur: this can be expressed more cleanly with the jtframe_edge module.

cores/wardner/hdl/jtwardner_video.v
asturur, line 1: this file breaks the repo conventions.
1 file per module.
so warner_tile and warner_obj should be files.
You are also missing the colmix file.

In general since you haven't done a decap of the tile engine or the obj engine, if those game capabilities fits into jtframe_scroll ( tile handler with scrolling ) and one of the jtframe objects generic modules, use those. you will just have tiny wrapper around them to adapt the data rather than hand made verilog implementation from mame of tile managers/sprite handlers.

If those have nothing special and we don't have a decap to prove the difference, the generic is just fine.

modules/jt32010/hdl/jt32010.v
```diff
@@ -0,0 +1,568 @@
...
+ *   ph1  latch the instruction word, advance PC
+ *   ph2  present the data-memory address
+ *   ph3  the data word has arrived: compute, and write back
+ *
+ * Multi-cycle instructions repeat the four phases. `mcyc` counts which
+ * machine cycle of the current instruction is running.
+ *
+ * `bio` follows MAME's bio() callback rather than the physical pin: drive it
+ * high to make BIOZ branch. The real BIO pin is active low.
+ */
```
asturur, line 28: This is mostly @jotego territory, but before adding a shared module there are some consideration to do. In primis the source/quality of the module. If it is a DSP generated by mame source code ( no datasheet, no decap no reverse engineering ) initially is ok to have it the game folder that introduces it as jtwardner_32010. Then at some point it can be elevated to shared module.

modules/jt32010/ver/cpu/regress.sh
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env bash
```
asturur, line 1: we do regression test in a different way, check how other modules do.
Ask ai to add test in a way that the CI can auto pick them up

asturur:
@memmerson i don't have merge approval rights, i commented on your code the comments i usually get from @jotego, the easy ones, the first round of objections on an ai core.

I m not a verilog expert so i may have miss many things.

cores/wardner/cfg/mem.yaml
asturur: Ask AI to use the pixel cen from the pll rather than the hand crafted one. i would give you more precise instructions if i was more expert.

memmerson:
> @memmerson i don't have merge approval rights, i commented on your code the comments i usually get from @jotego, the easy ones, the first round of objections on an ai core.
> 
> 
> 
> I m not a verilog expert so i may have miss many things.

Hey @asturur thanks for the direction, it's appreciated. I'll work my way through your suggestions tomorrow and over the weekend 👍

skutis:
This might be a better alternative: https://github.com/ika-musume/IKA32010

asturur:
> This might be a better alternative: https://github.com/ika-musume/IKA32010

@memmerson Yes please remove your implementation, vendor this one under modules, ask ai to create a jtframe wrapper for the module and cite to do the same work has been done for the taito c-chip. ( you can also check the old pr ) In this way correct licenses headers and file structures are used.  A module from ika/victor is definetely trustable as a shared module.

Jotego may still have something to deal with for the bsd license if is not gplv3 compatible

real-amano:
https://github.com/jotego/jtcores/tree/master/cores/ktiger/sch

Jotego has the Kyuukyouku Tiger schematics here. It is probably worth feeding them in the AI and hunt for obvious diacrepancies. And check which could be applicable for Wardner.

memmerson:
@asturur thanks, that was a genuinely useful first pass. Fourteen commits, roughly one per
comment, so each thread maps to a diff you can read on its own. Everything is
verified in simulation and the core has been run on a DE10-Nano and plays
correctly.

## Done

| Comment | Commit | |
|---|---|---|
| `plan.md` not needed in the repo | `99d4ca0` | Removed. Kept the vendored MAME sources — 71 other cores keep theirs the same way, so I assumed that part was fine. Say if not. |
| audio section belongs above sdram | `8af050b` | Reordered to clocks, audio, sdram, bram. The generated `jtwardner_game_sdram.v` and `mem_ports.inc` are byte-identical before and after. |
| the DSP ROM comment isn't useful | `8af050b` | Trimmed, and took the same pass over the other long comments in `mem.yaml`. |
| `cfg/msg` should say how the core was made | `6e99995` | Rewritten in the style of the recent cores. Compiles to 3072 bytes, exactly the 3 pages `JTFRAME_CREDITS` reserves, 15 rows spare. |
| no functions for one-liners | `8d10c54` | Two assigns alongside `cab_sys`, same expression. |
| `jtframe_edge` would be cleaner | `877d12f` | Done. Checked against the old latch over 200192 clocks of random `LVBL`/`int_en`, 0 mismatches. The only difference left is that `jtframe_edge` resets synchronously; the Z80 is held in the same reset, so it is not observable. |
| are those two assigns needed? | `8a9fc3e` | Both are, and `dsprom_addr` is not dead — `jtwardner_game_sdram.v` feeds it into the dsprom block RAM. The ports are genuinely narrower than their drivers, `[11:1]` against the DSP's 12-bit address and `[16:2]` against the tile engine's 16. The comment never said so, which is what made the line look wrong. Rewritten. |
| regression tests CI can pick up | `4a14249` | `modules/jt32010/ver/cpu` is now a jtframe unit test — `.simunit`, `gather.f`, `init.go`, `test.v`. Six cases, 24000 instructions against the C reference model. Ran it inside the `jotego/linter` container the way `xsimunit.sh` invokes it: exit 0. |
| one module per file | `703023d` | `jtwardner_video.v` split into `jtwardner_tile.v` and `jtwardner_obj.v`, and the mixer lifted into the `jtwardner_colmix.v` you said was missing. |
| board-level RAMs go in mem.yaml | `d041cb8`, `635b6a9` | All thirteen. Each `_lo`/`_hi` pair is now one 16-bit memory with byte write enables: `A[0]` picks the half for the Z80, and the DSP writes both at once while it holds the Z80 halted. No `jtframe_dual_ram` left in `jtwardner_main.v`. |
| use the jtframe tile handler | `7b75abd` | All three tile maps are `jtframe_scroll` now and `jtwardner_tile.v` is gone. |
| use a jtframe object module | `17f8bfd` | The line buffer is `jtframe_obj_buffer`. The scan and the draw are still the core's own — see the question at the end, which turned out to be more interesting than I expected. |

Still to do from your list: the pixel clock off the PLL with `pll_700` and
`SHIFT_180`. I have left it until the DSP question below is settled, because the
same fractional chain feeds `cen14`, which is the DSP's clock enable, and I would
rather not rebuild it twice.

## Verification

Nothing above is meant to change behaviour, and I tried to show that rather than
assert it.

- **Video frame bench: 42 of 42 random snapshots pixel-exact** — 30 sparse and 12
  dense at 512 overlapping sprites each, both screen-flip states, ROM latency 0
  to 4 — plus 8 of 8 named snapshots from real attract-mode frames.
- **Boot bench identical to the original code** throughout the RAM move: same
  milestones, same 253276 clocks halted by the DSP, same memory and port
  counters. For the RAM move I also diffed the snapshot *contents* — tx, bg, fg,
  palette, sprite RAM and the registers, all byte-identical — because the bench
  counters count writes and not what was written, and a swapped byte lane would
  pass them and still break the core.
- **jt32010 unit test**, 6 cases, 24000 instructions.
- **Verilator lint clean, Quartus 0 errors and 246 warnings**, which is the
  warning count the core had before any of this.
- **Run on a DE10-Nano and played.**

Resources moved a long way, almost all of it from the line buffer:

```
              before    after
ALMs          26246     15209      63% -> 36% of the device
registers     36801     23298
RAM blocks      142       144
```

Written by hand the sprite line buffer needed two write ports on one array,
which Quartus cannot map to block RAM, so it built 1024 thirteen-bit 3:1 muxes
out of logic. That was 28% of the core. `jtframe_obj_buffer` is the same
arrangement done properly and two M10Ks replaced 11103 ALMs. Worth knowing if
anyone else writes a line buffer by hand.

Going the other way, `jtframe_scroll` costs about 240 ALMs more than the
hand-written tile engine did, since it carries zoom and column scroll this core
does not use. That seems a fair trade for not maintaining a bespoke one.

Two things in `jtframe mem` that cost me time, neither of which errors:

1. A bram's **primary** port `dout:` is not honoured by the RAM instance — it
   keeps `<name>_dout`, while the generated game-module port takes the custom
   name and ends up undriven through an implicit 1-bit net. `dual_port.dout` is
   honoured. I worked around it by not overriding the primary `dout`.
2. `dual_port: { name: X }` silently reuses an existing bus named X. Naming one
   `snd` made it borrow the 15-bit SDRAM `snd_addr` and slice `[10:0]` off it.
   That happens to be right for this core, but only by luck.

## The DSP, and IKA32010

Happy to do this, and @skutis is probably right that a module from ika/Victor is
the better thing to depend on. Before I delete work, two things:

1. **The licence.** You flagged it yourself — if BSD into a GPL-3.0 repo is a
   problem, that is @jotego's call and it decides the whole question. Is there a
   precedent I should follow?
2. **Verification.** The only evidence I have that the DSP behaves is that it
   matches a transcription of MAME's `tms320c1x.cpp` instruction for
   instruction. Swapping the implementation discards that unless the new one
   passes the same traces. The test I just added is the instrument for checking
   it, so I would rather point it at IKA32010 and see than swap on trust. Does
   that sound reasonable, or is the module trusted enough that this is wasted
   effort?

If the answer to (1) is no, your earlier suggestion stands and I will move ours
into the core folder as `jtwardner_32010`.

## The sprite engine, and what the schematics turned up

@real-amano - thank you, that pointer was worth more than the rest of the day.
It settled three things, in both directions.

I had been about to tell you the sprite scan and draw have to stay bespoke
because `jtframe_objdraw_gate` cannot express this core's priority rule: MAME
marks every pixel a sprite touches so that later sprites are never masked there,
and reproducing that needs the line buffer's existing value at the pixel being
written, while `buf_pred` is the value the module proposes to write. That is
still true as a fact about the API. But the schematics suggest the rule should
not be there at all.

### 1. The priority PROM, where MAME looks wrong

`Prio.kicad_sch` resolves layer priority with a 32x8 PROM. Tracing it, the five
address lines are:

```
A4 = GFXA[2:0] != 0      text layer opaque   (three planes, not four)
A3 = OBJPXL9             sprite priority bit 1
A2 = OBJPXL8             sprite priority bit 0
A1 = OBJPXL[3:0] != 0    sprite opaque
A0 = GFXC[3:0] != 0      foreground opaque
```

No background-opaque input, because the background is always opaque. Reading the
real PROM out of the ROM set - `82s123.f1`, crc32 `4fb5df2a`, the same part as
Twin Cobra's `82s123.b24` - and decoding it through that map gives a completely
coherent table, which is itself decent evidence the trace is right:

| sprite priority | over background | over foreground | over text |
|---|---|---|---|
| 00 | never drawn | — | — |
| 01 | yes | no | no |
| 10 | yes | yes | no |
| 11 | yes | yes | yes |

That is exactly what `jtwardner_colmix` already does, with one extra term:
`obj_multi`, our copy of MAME's touched-pixel marking. The PROM has five address
lines and all five are accounted for; none is a touched bit. Nor could it be
stored - from `linebuffer.kicad_sch` the sprite line buffer is `OBJPXL[0..11]`,
`PXLA[0..11]`, `PXLB[0..11]`, two banks of 1024 by twelve bits in six MBM2148s.
Twelve bits is exactly `{priority[1:0], colour[5:0], pen[3:0]}` with nothing
spare. MAME's "priority 31" does not fit in two bits; it comes from MAME's own
priority-bitmap machinery, not from this board. MAME does not claim otherwise -
its driver loads three PROMs it never uses and annotates this one
`// tile to sprite priority ??`.

The cost was already measured here: `render_ref.py` has carried a `--no-pri31`
flag from the start, and dropping the term changes **5366 pixels of a dense fuzz
frame and 4 pixels of one real gameplay frame**. So it shows whenever sprites
overlap.

I have not changed it. The trace is of Kyuukyouku Tiger's board and Wardner is
TP-009 on Taito B25 - same PROM, same contents, but I have not seen a Wardner
schematic - and switching means `render_ref.py`, which is deliberately kept a
transcription of MAME so the frame diff checks the RTL rather than an opinion,
would have to deviate from MAME on purpose. That is your call, not mine.

If anyone has a Wardner or Twin Cobra board: two overlapping sprites over
foreground terrain, and see which wins.

### 2. Two things where MAME is right and this core doubted it

`ver/video/README.md` has been carrying two flip behaviours as "matched to MAME
but unlikely to be what the board does". Both were wrong to doubt, and `64d6e6f`
fixes the documentation.

**Sprites do not flip.** The top sheet routes `FLIP` to `gfxa`, `memorias` and
`misc` only. `object` has nineteen sheet pins and `FLIP` is not among them. A
flipped cabinet really does show mirrored tilemaps over unmirrored sprites.

**The flipped view is a clean 180 degree mirror after all.** It had looked
displaced by 79 pixels horizontally and 213 vertically. That was an artefact of
how it was measured - forcing `flip=1` on a frame captured unflipped keeps scroll
values the software would never write in that state. Running the boot bench with
the Flip Screen DIP actually set, the game writes **scroll 79 lower horizontally
and 213 lower vertically**, cancelling MAME's second offsets exactly:

```
clean mirror = 55 + scrollx + (319 - screen_x)  = 374 + scrollx - screen_x
MAME flipped = 453 + (scrollx - 79) - screen_x  = 374 + scrollx - screen_x
```

Both runs reach 16000 ms with byte-identical text, background, foreground,
palette and sprite RAM - only the scroll registers and the flip bit differ - so
the frames are the same screen. Rendered without sprites, the flipped frame
matches the control rotated 180 degrees on **all 76800 pixels**. So
`set_scrolldx(-55, -134)` and `set_scrolldy(-30, -243)` are correct.

While tracing this I also found that `FLIP`'s job in hardware is to reverse the
74LS299 shift direction so each tile is mirrored within itself - which is what
`XOR_HFLIP` does in `jtframe_tilemap`, and what I had to set empirically when the
tile layers moved to `jtframe_scroll`. Pleasing to find the board had been doing
it all along. `jtframe_scroll` also mirrors by XOR of the position, which is what
the 74LS86 gates in `memorias` do, where MAME subtracts from a constant - so
moving these layers on to the generic took the core structurally closer to the
board, not further.

### 3. One PROM label in MAME is a wrong guess

`82s123.c6`, annotated `// sprite attribute (flip/position) ??`, sits in
`objctrl` and is an address decoder: `A0..A4 = RAAD8..RAAD12`, outputs driving
`RA2..RA4` to select among the five 6264s, gated by `R~{w}`. Nothing to do with
sprites - consistent with `FLIP` never reaching them.

### Where that leaves the sprite engine

**If the touched-pixel rule should go, the question reopens.** Without
`obj_multi` there may be nothing stopping the scan and draw moving on to the
jtframe object modules after all. I would rather settle that than argue the core
is special on the strength of a quirk that may not be real.

The netlist tracer I used is a short Python script and every claim above is
reproducible with it; happy to contribute it if it would be useful to anyone else
reading these schematics.

cores/wardner/hdl/jtwardner_game.v
```diff
@@ -0,0 +1,275 @@
...
+    .host_addr  ( dsp_addr      ),
+    .host_sel   ( dsp_sel       ),
+    .host_dout  ( dsp_dout      ),
+    .host_din   ( dsp_din       ),
+    .host_we    ( dsp_we        ),
+
+    .rom_addr   ( dsp_rom_addr  ),
+    .rom_data   ( dsprom_data   ),
+
+    .dbg_bio    (               ),  .dbg_exec   (   ),
```
jotego, line 182: Let's remove the debug ports or at least simplify them

cores/wardner/hdl/jtwardner_game.v
```diff
@@ -0,0 +1,275 @@
...
+    .video_on   ( video_on      ),
+
+
+    .dipsw_a    ( dipsw_a       ),
+    .dipsw_b    ( dipsw_b       ),
+    .joy1       ( cab_joy1      ),
+    .joy2       ( cab_joy2      ),
+    .cab_sys    ( cab_sys       ),
+
+    .dbg_iowr   (               ),
```
jotego, line 156: Remove debug ports. Tell the AI to look at the test.fst waveform traces instead of producing text output for debugging

cores/wardner/hdl/jtwardner_game.v
```diff
@@ -0,0 +1,275 @@
...
+// The Toaplan cabinet ports read active high, JTFRAME's inputs are active low.
+// SYSTEM (wardner.cpp): 0 service, 1 tilt, 2 test, 3 coin1, 4 coin2,
+// 5 start1, 6 start2, 7 vertical blanking - and the main CPU substitutes
+// blanking for bit 7 itself, so it is left at zero here.
+//
+// dip_test and tilt are active low like the rest: jtframe_dip.v says of
+// dip_test "assumes it is always active low" and drives it to zero to assert
+// test in simulation, and the CPS cores tie an unused tilt to 1'b1. Passing
+// either through uninverted asserts TEST and TILT on the board permanently.
+assign cab_sys = { 1'b0, ~cab_1p[1], ~cab_1p[0], ~coin[1], ~coin[0],
```
jotego, line 75: Move cabinet related items inside the main module

cores/wardner/cfg/mem.yaml
jotego: This pxl_CEN will create artifacts. Clock enable for the pixels must be an integer division, not a fractional one. Otherwise, the jitter will be visible. 
As @asturur said, set a 7MHz PLL for this. If that creates problems, let me know.

cores/wardner/hdl/jtwardner_game.v
```diff
@@ -0,0 +1,275 @@
...
+
+assign { dipsw_b, dipsw_a } = dipsw[15:0];
+`ifdef JTFRAME_IOCTL_RD
+    assign pal_vaddr = ioctl_ram ? ioctl_addr[11:1] : pal_vaddr_vid;
+    assign ioctl_din = ioctl_addr[0] ? pal_vq[15:8] : pal_vq[7:0];
+`else
+    assign pal_vaddr = pal_vaddr_vid;
+`endif
+assign dip_flip = flip;
+assign pxl_cen  = cen7;
```
jotego, line 63: See the comment above.

cores/wardner/hdl/jtwardner_game.v
```diff
@@ -0,0 +1,275 @@
...
+// the DSP addresses 4096 words; the mask ROM region holds 2048
+wire [11:0] dsp_rom_addr;
+wire        obj_ovf, sample;
+
+// the sprite generator asks for a 16-pixel row as two 32-bit halves, so the
+// bus address is {sprite, row, half}; the tile maps ask for {tile, row}
+wire [15:0] fg_full_addr;
+
+assign { dipsw_b, dipsw_a } = dipsw[15:0];
+`ifdef JTFRAME_IOCTL_RD
```
jotego, line 56: This should be moved somewhere else. Probably in the colmix module

cores/wardner/hdl/jtwardner_game.v
```diff
@@ -0,0 +1,275 @@
+/* SPDX-FileCopyrightText: 2026 Marc Emmerson
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ *
+ * Wardner (Toaplan TP-009 / Taito B25, 1987) top level.
```
jotego, line 4: Eventually remove as many comments as possible.

cores/wardner/hdl/jtwardner_main.v
```diff
@@ -0,0 +1,314 @@
...
+    // program ROM
+    output reg [17:0] rom_addr,
+    input      [ 7:0] rom_data,
+    output reg        rom_cs,
+    input             rom_ok,
+
+    // DSP subsystem
+    output            dsp_on,         // run bit: LS259 "coinlatch" bit 0
+    input             dsp_halt,       // the DSP is holding this CPU
+    input      [12:0] dsp_addr,
```
jotego, line 55: For 16-bit systems, start counting the LSB for addresses at 1 instead of 0

cores/wardner/hdl/jtwardner_main.v
```diff
@@ -0,0 +1,314 @@
...
+    input      [ 7:0] dipsw_a, dipsw_b, joy1, joy2, cab_sys,
+
+    // observation, for the bench
+    output            dbg_iowr, dbg_iord,
+    output     [ 7:0] dbg_port, dbg_data,
+    output     [15:0] dbg_pc_addr,
+    output            dbg_m1
+);
+
+// ------------------------------------------------------------------ the CPU
```
jotego, line 103: Remove separators like this

cores/wardner/hdl/jtwardner_main.v
```diff
@@ -0,0 +1,314 @@
...
+assign dbg_data    = cpu_dout;
+assign dbg_pc_addr = A;
+assign dbg_m1      = ~m1_n;
+
+// ------------------------------------------------------------------ banking
+reg [2:0] bank;
+reg       ram_view;                 // 1 = the RAMs answer reads at 8000-FFFF
+
+// ---------------------------------------------------------- memory decoding
+wire in_rom0 = mreq && A < 16'h7000;
```
jotego, line 131: Use a single always block to make the memory address decoding and call the decoded signals with a _cs prefix.

cores/wardner/hdl/jtwardner_main.v
```diff
@@ -0,0 +1,314 @@
...
+
+// ------------------------------------------------------------ I/O registers
+wire io_wr = iorq & wr;
+wire io_rd = iorq & rd;
+
+reg [7:0] coin_ctl;                 // LS259 at port 5A
+reg       int_en;                   // LS259 at port 5C, bit 2
+assign dsp_on = coin_ctl[0];
+
+always @(posedge clk, posedge rst) begin
```
jotego, line 192: Prefer using synchronous reset

cores/wardner/hdl/jtwardner_main.v
```diff
@@ -0,0 +1,314 @@
...
+        txoffs <= 0; bgoffs <= 0; fgoffs <= 0;
+        tx_scrx<=0; tx_scry<=0; bg_scrx<=0; bg_scry<=0; fg_scrx<=0; fg_scry<=0;
+        flip <= 0; bg_bank <= 0; fg_bank <= 0; video_on <= 0;
+        int_en <= 0;
+        coin_ctl <= 0;
+    end else begin
+        tx_lo_we <= 0; tx_hi_we <= 0;
+        bg_lo_we <= 0; bg_hi_we <= 0;
+        fg_lo_we <= 0; fg_hi_we <= 0;
+        if( cpu_cen && io_wr ) begin
```
jotego, line 204: This may be done better using a cfg/mmr.yaml file instead for readability and scene simulation support.

cores/wardner/hdl/jtwardner_main.v
```diff
@@ -0,0 +1,314 @@
...
+
+// ------------------------------------------------------------- read multiplex
+always @* begin
+    cpu_din = 8'hff;
+    if( rom_cs )      cpu_din = rom_data;
+    else if( in_work )               cpu_din = A[0] ? work_dout[15:8] : work_dout[7:0];
+    else if( in_obj && ram_view )    cpu_din = A[0] ? objram_dout[15:8]  : objram_dout[7:0];
+    else if( in_pal && ram_view )    cpu_din = A[0] ? pal_dout[15:8]  : pal_dout[7:0];
+    else if( in_snd && ram_view )    cpu_din = shared_dout;
+    else if( io_rd ) case( port )
```
jotego, line 276: Register the cabinet input separately to ease timing

cores/wardner/hdl/jtwardner_main.v
```diff
@@ -0,0 +1,314 @@
...
+    .clk    ( clk       ),
+    .edgeof ( ~LVBL     ),
+    .clr    ( ~int_en   ),
+    .q      ( irq_n     )
+);
+
+// ------------------------------------------------------------- read multiplex
+always @* begin
+    cpu_din = 8'hff;
+    if( rom_cs )      cpu_din = rom_data;
```
jotego, line 271: Use a case or a multi-line ternary statement instead of the cascade of if statements.

cores/wardner/hdl/jtwardner_main.v
```diff
@@ -0,0 +1,314 @@
...
+        8'h61: cpu_din = txram_dout[15:8];
+        8'h62: cpu_din = bgram_dout[ 7:0];
+        8'h63: cpu_din = bgram_dout[15:8];
+        8'h64: cpu_din = fgram_dout[ 7:0];
+        8'h65: cpu_din = fgram_dout[15:8];
+        default: cpu_din = 8'hff;
+    endcase
+end
+
+T80s u_cpu(
```
jotego, line 292: Instantiate a JTFRAME wrapper for the CPU

cores/wardner/hdl/jtwardner_obj.v
```diff
@@ -0,0 +1,233 @@
...
+    output reg [15:0] rom_addr,
+    input      [31:0] rom_data,
+    output reg        rom_cs,
+    input             rom_ok,
+    output reg [12:0] pxl,
+    output reg        ovf
+);
+
+// ---- frame copy of the object RAM, 2048 clocks from the fall of LVBL
+reg  [15:0] frame[0:2047];
```
jotego, line 35: Either use a JTFRAME RAM module or a BRAM entry in mem.yaml.

cores/wardner/hdl/jtwardner_obj.v
```diff
@@ -0,0 +1,233 @@
...
+reg  [15:0] frame_q;
+reg  [10:0] cp, cp_addr;
+reg         copying, LVBL_l, we1, we2;
+reg  [10:0] frd_addr;
+
+always @(posedge clk) begin
```
jotego, line 41: This will mostly go away once you use a JTFRAME ram module

cores/wardner/hdl/jtwardner_obj.v
```diff
@@ -0,0 +1,233 @@
...
+
+always @(posedge clk, posedge rst) begin
+    if( rst ) begin
+        copying <= 0; cp <= 0; LVBL_l <= 1;
+    end else begin
+        LVBL_l <= LVBL;
+        if( LVBL_l && !LVBL ) begin
+            copying <= 1;
+            cp      <= 0;
+        end else if( copying ) begin
```
jotego, line 58: This is not realistic. Take the copying speed down to something the actual board can do. Probably 1/4 of the video clock will be close to reality.

cores/wardner/hdl/jtwardner_obj.v
```diff
@@ -0,0 +1,233 @@
...
+);
+
+always @(posedge clk, posedge rst) begin
+    if( rst ) pxl <= 0; else pxl <= lbuf_q;
+end
+
+// ---- renderer. Scans entry 511 down to 0 reading the y word, one entry a
+//      clock through a small pipeline; on a hit it fetches attr, code and x,
+//      two ROM words, and writes the 16 pixels one a clock.
+reg  [ 3:0] st;
```
jotego, line 100: Separate the scanning function from the drawing and delegate the drawing to the standard obj drawing modules in JTFRAME

cores/wardner/hdl/jtwardner_sound.v
```diff
@@ -0,0 +1,146 @@
...
+wire [ 7:0] cpu_dout;
+reg  [ 7:0] cpu_din;
+
+wire rd   = ~rd_n;
+wire wr   = ~wr_n;
+wire mreq = ~mreq_n & rfsh_n;
+wire iorq = ~iorq_n & m1_n;
+
+// ------------------------------------------------------------------ decoding
+wire in_rom  = mreq && !A[15];                              // 0000-7FFF
```
jotego, line 61: Same as for the main CPU address decoder

cores/wardner/hdl/jtwardner_sound.v
```diff
@@ -0,0 +1,146 @@
...
+assign rom_addr = A[14:0];
+
+assign shr_addr = A[10:0];
+assign shr_dout = cpu_dout;
+assign shr_we   = in_shr & wr;
+
+// ---------------------------------------------------------------- local RAM
+wire [7:0] ram1_q, ram2_q;
+
+jtframe_dual_ram #(.AW(7),.DW(8)) u_ram1(
```
jotego, line 76: Move the shared memory to mem.yaml

cores/wardner/hdl/jtwardner_sound.v
```diff
@@ -0,0 +1,146 @@
...
+jtframe_dual_ram #(.AW(11),.DW(8)) u_ram2(
+    .clk0(clk),.data0(cpu_dout),.addr0(A[10:0]),.we0(in_ram2 & wr),.q0(ram2_q),
+    .clk1(clk),.data1(8'd0),.addr1(11'd0),.we1(1'b0),.q1() );
+
+// ------------------------------------------------------------------- YM3812
+wire       fm_cs = iorq && A[7:1] == 7'd0;      // ports 00 and 01
+wire [7:0] fm_dout;
+wire       fm_irq_n;
+
+assign dbg_fmwr   = fm_cs & wr;
```
jotego, line 89: Remove these debug signals and scan the test.fst file directly

cores/wardner/hdl/jtwardner_sound.v
```diff
@@ -0,0 +1,146 @@
...
+        in_ram2: cpu_din = ram2_q;
+        in_shr:  cpu_din = shr_din;
+        fm_cs:   cpu_din = fm_dout;
+        default:;
+    endcase
+end
+
+wire wait_n = ~(rom_cs & ~rom_ok);
+
+T80s u_cpu(
```
jotego, line 124: Instantiate a JTFRAME wrapper instead

cores/wardner/hdl/jtwardner_video.v
```diff
@@ -0,0 +1,208 @@
...
+wire [8:0] bg_scry_eff = flip ? bg_scry[8:0] - 9'd29 : bg_scry[8:0] + 9'd30;
+wire [11:0] bg_vaddr_lo;
+wire [31:0] bg_data_rev = { bg_data[7:0], bg_data[15:8], bg_data[23:16], bg_data[31:24] };
+
+assign bg_vaddr = { bg_bank, bg_vaddr_lo };
+
+jtframe_scroll #(
+    .SIZE(8), .CW(12), .VA(12), .PW(8), .MAP_HW(9), .MAP_VW(9),
+    .FLIP_HW(9), .FLIP_VW(9), .HJUMP(1), .XOR_HFLIP(1)
+) u_bg(
```
jotego, line 125: Write one port per line

cores/wardner/ver/dsp/mutate.sh
jotego, line 1: Make verilog tests following the simunit convention used in the repo. Keep ad-hoc tests to a minimum.

cores/wardner/ver/game/README.md
```diff
@@ -0,0 +1,161 @@
+# Wardner whole-core bench
+
+Runs `jtwardner_game.v` under jtcores' own `jtsim` harness, with the real
+SDRAM controller and the C++ SDRAM model, so the thing being tested is the
+core as it will be built rather than a hand-wired bench.
+
+## Building the ROM
```
jotego, line 7: You build the rom using `jtframe mra`. Its better if you write the README manually or keep it very short. 
I will eventually delete all AI generated text that people may read and write things normally.

cores/wardner/ver/game/mkrom.py
jotego, line 1: Delete this file

cores/wardner/ver/game/raw2png.py
jotego, line 1: Delete this file

cores/wardner/ver/main/crtc_check.py
jotego, line 1: Delete all python tests that are testing the verilog.

cores/wardner/ver/main/mkrom.py
jotego, line 1: Delete this file

jotego (changes requested):
Thank you for submitting this work. 

The verification does not follow typical digital engineering approaches or this repo's standard files. I think we can just wipe it out completely. Maybe add specific simunit tests for worrying aspects. 

The video module is a bit of a small mess too but not too bad. The worst problem is the fractional pixel clock enable. 

The object module should be kept to simply the scanner and reuse JTFRAME features for the rest. 

There might be a bus request/grant protocol for the object DMA that is not implemented at the moment. Do you have schematics for this board?

Does this compile for pocket and the other platforms too? Please try at least compiling for the three big ones: mister, pocket and sidi128

Thanks!

memmerson:
Hi @jotego thank you for your review, I will work through the comments today. 

Whilst I'm doing that, have you any thoughts on the DSP? We could use [IKA32010](https://github.com/ika-musume/IKA32010) (BSD licensed) as suggested by skutis and asturur. I've already made another branch locally using it, and it's working, so I could merge back into this branch if you're happy with that?

jotego:
Thanks, about the MCU, I think IKA's module is a better choice than spawning an AI-generated one. 

I would normally use a git submodule for it but because the repository contains PDF files that are copyrighted we should just copy the hdl files and the license to a folder inside jtframe/hdl/cpu/new-folder. Create a files.yaml for it so it can be referenced with a single line by the core. If there is an associated PROM, it may be better to create a wrapper that instantiates the MCU with the PROM and renames the ports to more familiar names in the framework.

asturur:
@memmerson look how in this PR https://github.com/jotego/jtcores/pull/1486 the jttc0030 folder under modules is organized. in that PR i took an ika work and made a module out of it.

memmerson:
@jotego @asturur thanks both, the review is addressed and the branch is updated. Runs on MiSTer hardware.

- **Verification:** the old benches are gone. Two simunit tests replace them: `ver/dsp_handshake` and `ver/obj_order`.
- **Pixel clock:** now an integer division, no fractional cen. The base clock is 56 MHz on MiSTer (`jtframe_pll7000` + `JTFRAME_180SHIFT`). On SiDi128 it's 49 MHz, from a new `jtframe_pll6125`, because the SiDi128 game PLL doesn't lock at 56.
- **CPUs:** jtframe wrappers, `_cs` decoding in one always block, `mmr.yaml` for the scroll registers, synchronous reset, registered cabinet inputs moved into main, debug ports removed. The shared and sound RAMs are in `mem.yaml`.
- **Objects:** the core only scans; drawing is `jtframe_objdraw`. The RAM copy runs at 1/4 of the pixel clock.
- **Video:** one port per line, palette readback in colmix, comments trimmed.
- **DSP:** IKA32010 in `jtframe/hdl/cpu/ika32010`, used through `jtframe_tms32010`. @asturur it follows the jttc0030cmd layout; only the folder differs, since jotego asked for `jtframe/hdl/cpu`. Three local changes, listed in its README:
  - the disassembler `define` commented out, so Quartus can compile it
  - a CALL fix (RET was returning one word early)
  - lint pragmas

- **Platforms:** MiSTer and SiDi128 compile and meet timing. I can't build Pocket because the `jotego/pocket` submodule is private. Could you approve the CI runs?
- **Bus request/grant:** there are no Wardner schematics, but the Kyuukyoku Tiger sheets show it.
  - The DSP takes the 68000 bus with BR/BG/BGACK, and BGACK drives the DSP's BIO.
  - The main CPU reaches the sound RAM through the Z80's BUSRQ/BUSACK.
  - The sprite copy uses no bus request; it's done by counters and buffers.

  The core follows MAME: an immediate CPU halt for the DSP and a dual-port shared RAM. The difference is only timing, but I can switch the main Z80 to BUSRQ/BUSAK if you prefer.

memmerson:
Platforms: MiSTer, SiDi128 and Pocket all compile and meet timing on the current branch (Pocket 5,663 ALMs, 31%). @jotego thanks for making the Pocket files public.

PR #1598, opened 2026-09-11
Grad3: Fixes #1583

### LINKED ISSUE #1583: grad3:Overall game performance is too slow under default clock setting
### Description
In the latest build (`e00a9c6`), the overall game speed under default clock settings (Turbo OFF) is significantly slower than the original arcade PCB.

While this slowdown affects the entire game, it is especially severe in Stage 4 (3D stage). As shown in the comparison video, the original PCB clears the 3D maze segment in approximately 47 seconds, whereas on the core, the BGM loops entirely and the player is still only about halfway through the section.

### Video Comparison

- **Left:** jtcores@e00a9c6 (Default clock) — Still halfway through after the BGM loops
- **Right:** Original PCB — Clears the section in ~47 seconds

https://github.com/user-attachments/assets/6091fda9-98dc-4798-b3fa-c8f4c32b5610

niknakniknak:
Fixes #1583

A fix for the slow stage 4.  The sub CPU's level 1 and level 2 are on the wrong VBLK edges. The sub now feeds ~LVBL into jtgrad3_int, so it takes level 1 at the start of blanking and level 2 at the end. The master and jtgrad3_int are unchanged. Stage 4 is 47s again, same as the PCB.

This doesn't follow the schematics as drawn. on the sub sheet they have level 1 (L11A) on ~VBLK and level 2 (L11B) on VBLK through K11E. The master sheet matches MAME, but the sub sheet is the other way round from it: MAME has the sub's level 1 at line 240 and level 2 at line 16. With this change it lines up with MAME exactly (which works at tehe correct speed here) My private core also followed the schematics and gets the same 77 seconds, so the schematics might have the two swapped on the sub?

@skutis , if you still has the board,

On the sub, is L11A pin 3 really on ~VBLK, and L11B pin 11 on K11 pin 10? Or are they the other way round? curious to know

I've tried these fixes for this..

F22, the new cache and turbo ruled out by bisect.
Flipped the CPU triggered interrupt polarity: 77s.
Gated main CPU sub_irq with dws: 77s.
IPL priority encoder in jtgrad3_int: 77s.
Reverted 7b735832e on current master: 47s.
Putting back the sprite timed level 2 (sub_irq2) removed in 7b735832e: 47s. (not keen)
LVBL polarity reversed in jtgrad3_int, so both CPUs: 77s.
Keeping 7b735832e and swapping the sub's level 1 and 2 only: 47s. (winner winner)

Full play through tests fine.

### DIFF
--- a/cores/grad3/hdl/jtgrad3_sub.v
+++ b/cores/grad3/hdl/jtgrad3_sub.v
@@ -152,7 +152,7 @@ jtframe_okdly #(.W(3)) u_okdly(
 jtgrad3_int u_int(
     .rst      ( rst_cpu          ),
     .clk      ( clk              ),
-    .LVBL     ( LVBL             ),
+    .LVBL     ( ~LVBL            ),
     .cpu_trig ( cpu_trig         ),
     .din      ( cpu_dout[10:8]   ),
     .wr       ( irq_mask_cs      ),


niknakniknak:
good bot.

skutis:
I'm sorry it is correct in the sch.

niknakniknak:
Thanks @skutis  Odd then, but the swap gets the PCB timing so I'll leave it there.

jotego:
We cannot take this _fix_. This does not agree with the schematics and it makes the main and sub CPUs inconsistent. On the current schematics, both the main and the sub CPUs have the same interrupts levels:

- IPL2: triggered by the other CPU
- IPL1: triggered when entering blanking
- IPL0: triggered when leaving blanking

This patch changes the behavior only for the sub CPU. 

@skutis, could it be that the IPL wiring is indeed different for each CPU?

skutis:
I'll check everything tonight. Thanks both of you.

skutis:
I did not find any discrepancies in the schematics.

- My suspicion is the bus arbitration, could it be the cause? I mean mame does not implement it, it has to have some other way of granting shared RAM access deterministically.
  - ~{IORQ'} is missing from the force enable of master reg, also force disable of sub CPU.
- There are no vertical counters exposed on PCB to check VBLK against. Some other reference maybe?
- Otherwise, what is next thing to check?

niknakniknak:
Thanks @skutis for checking.

The arbiter, F22, reverting 7b735832e with it kept in still gets 47s, so I don't think that's this one. The ~IORQ' force on F22 I'll put in a separate PR along with F24, which isn't modelled yet.

I've found a different fix that leaves the interrupts as they are on the schematics: running the 051960 DMA at its original speed (dma_cen) for grad3 only gives 47s on stage 4 with no flicker (like whats happenign in crimfght). See PR #1604

jotego:
The DMA speed fix makes sense. The odd game for this is Crime Fighters and it may be due to the CPU being slightly faster or slower as we haven't measured the clock cycles per instruction yet. 
I close this PR and I will take the other one.

PR #1599, opened 2026-09-11
Grad3: Fixes #1585

### LINKED ISSUE #1585: grad3: Incorrect screen flashing effect during Mega Crash
### Description
In the latest build (`e00a9c6`), the full-screen flashing effect triggered by Mega Crash is not rendered correctly compared to the original arcade PCB.

- **Original PCB:** The entire background flashes completely white.
- **JT Core:** The space background layer does not turn fully white, appearing semi-transparent or darkened instead.

### Comparison
- **Left:** jtcores@e00a9c6 — Space background is not flashing fully white
- **Right:** Original PCB — Clears with a full white flash

Image

https://github.com/user-attachments/assets/5428b761-de28-4606-b192-f4ef59d23927

niknakniknak:
Fixes #1585

The Mega Crush flash writes white to palette 0x400, the backdrop. The PROM (J28 O3) sets the dim for every pixel with no layer on it, so the core halves it and the flash comes out mid grey. The PCB and MAME show white. This stops the backdrop being dimmed. Sprite shadows are untouched.

0x400 is black everywhere else in the game (checked in MAME through attract, all 12 stages and save states) so the flash is the only place this shows, it never draws a shadow either, I can see logged. only writes at 0 on boot. 

This doesn't follow the schematics as drawn, they run the dim (H26 pin 12 to C22A/B/C, 100R pull downs on the 052535 outputs) on the bcakdrop too. Going by the 052535 sheet that pull down would leave somewhere between a third and two thrids of the brightness depending on the monitor, not white, and the PCB flash is white. So the schematics might have an error there maybe? unless im reading it wrong.

@skutis 

Does H26 pin 12 really go to C22 pins 1, 3 and 5?
With a scope on C22 pin 1, is it low on the black background?
Curious.

I've tried these:

Zeroing the palette address when no layer is on (~CDEN): still grey.
Not dimming the backdrop: white, same as the PCB. (winner winner)

Full play through tests fine

### DIFF
--- a/cores/grad3/hdl/jtgrad3_colmix.v
+++ b/cores/grad3/hdl/jtgrad3_colmix.v
@@ -72,7 +72,7 @@ always @(posedge clk) begin
         rgb <= 0;
     end else if(pxl_cen) begin
         pal_rd_addr <= pal_addr;
-        shl <= shd;
+        shl <= shd | en_b;
         rgb <= shl ? palrd_dout[14:0] : dim(palrd_dout[14:0]);
     end
 end


skutis:
- Does H26 pin 12 really go to C22 pins 1, 3 and 5?
- With a scope on C22 pin 1, is it low on the black background?
Yes all this is true

niknakniknak:
Thanks @skutis then all I can think is that it just tames a hot signal for CRT's. Fix works anyway.

jotego:
It doesn't quite agree with the schematics as the dimming seems to always operate regardles of the /OE signal. Anyway, this is a small thing, I'm ok with the change.

niknakniknak:
thanks for merging

PR #1600, opened 2026-09-11
Harier: Fix Enduro PCM sample rate on YM2151

niknakniknak:
YM2151 versions of Enduro racer were playing an octave low, the PCM clock is now 8 MHz, giving a 31.25 kHz sample rate. sounds correct now.


jotego:
8MHz is too fast. That chip cannot work at that speed. The error is somewhere else.

niknakniknak:
t's the same 315-5218 and jtoutrun_pcm as Out Run, which already runs it from 8M; at 4M Enduro's samples play an octave low.

PR #1603, opened 2026-09-12
harier: Adds hang-on

niknakniknak:
adds Hang-On to harier: main and sub CPU map and clock, inputs and pedals, road control bit 9, SHADER shadow/hilight and the Hang-On sprite format.

Sound needs the T80 EI/NMI fix, PR #1602 without it the sound fails fairly spectacularly after a short, random time during play

### DIFF
--- a/cores/harier/hdl/jtharier_dtack_cen.v
+++ b/cores/harier/hdl/jtharier_dtack_cen.v
@@ -5,6 +5,7 @@
 module jtharier_dtack_cen(
     input         rst,
     input         clk,
+    input         hangon,
     output        cpu_cen,
     output        cpu_cenb,
     input         UDSn, LDSn,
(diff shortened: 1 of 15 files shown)


jotego:
The different width of the sprite data for Hang-On is suspicious. Can you investigate whether it is the same sprite logic as S16A?

niknakniknak:
Yes, it's the same logic, Hang on and space harrier both use the S16A 315-5011 315-5012 pair. The width comes from the ROM board, hangon uses two ROMs per word, Space Harrier four. Unlike S16A, Hang-On has zoom and its bank number is at +2.
Would you rather it used jts16_obj instead of jtharier_obj?

jotego:
If it is the same logic, then it is possible to keep reading the data with the same width for both games, regardless of whether it is stored in 2 or 4 chips. The data can be interleaved in many ways in order to make the two cases identical from the point of view of the core logic. This should be done in mame2mra.toml

The files in the cfg folder are there to simplify the verilog logic as much as possible. This is one of those cases.

Let me know if you cannot find a way of doing it and I will try myself then. Sometimes, a new feature is needed in the cfg files to support a simplification.

niknakniknak:
Done. mame2mra now repeats each hang on16-bit word to fill a 32-bit word, The only hangon difference left in jtharier_obj_draw is 4 pixels per word instead of 8. Tested, looks fine.

niknakniknak:
Does this need anything else from me?

cores/harier/hdl/jtharier_dtack_cen.v
```diff
@@ -20,6 +21,9 @@ module jtharier_dtack_cen(
     output  [15:0] fworst  // average cpu_cen frequency in kHz
 );
 
+wire [8:0] num = hangon ?  9'd1 :  9'd173;
```
jotego, line 24: It would be nicer for timing closure to register the num and den signals.

PR #1605, opened 2026-09-13
grad3: correctness additions

niknakniknak:
Brings the grad3 bus arbitration closer to the schematics.

F22: ~IORQ' forces the video bus grant to the main CPU (F22B ~R, from H22).
F22: bus requests come from the AS gated address decode, not DS (B23 and H13 into F22 D).
F22: grant synced to ~P1H' then ~CLK6M before DTACK (E22A/D23B main, K12B/J18B sub), sub ~OBJCS synced to P1H (K12A).
F24: shared RAM arbiter added, same circuit as F22 (F24, DTACK via D24/F25).
Arbitration waits passed as bus_legit so they aren't recovered.

Tested with #1604  stage 4 - 47s fullg ame tested


jotego:
How do you test this? Is the AI running a simulation? How many frames? Do you use a .cab file as input? How is the simulation output evaluated?

jotego:
The bus arbitration was the cause of the random deaths. Changing it again may bring back the problem. How are you testing that it is not happening again?

niknakniknak:
I build cores and play on a mister simming for the random death would be pointless and slow

niknakniknak:
The iorq was mentioned in the other pr as missing I added that. The rest were found after with ai going through the schems. And myself verifying the findings. I build cores and test physically with each addition. Leave it for now.

PR #1610, opened 2026-09-16
Outrun: Adds gears QOL

niknakniknak:
UPDATED BELOW

Adds 2 extra dips for OR and TOR

Gear button, Hold for low is default as is in the core now, and adds toggle. 
Gear Indicator, adds an on screen in game LOW/HI indicator based on the saturn sega ages version.
Fully tested on Mister.

OR
TOR

Note I had help from Opus 5.2 for this. Thought i'd offer it here.


jotego:
The toggle function is most welcome. It has been asked for in the past.

For the text display, it makes the logic implemented in the core _ugly_. It does not represent the real hardware and it is not a nice way of implement the feature either.

Would you mind patching the software for that instead? Sure the LLM can find where the gear is processed and then write the HI/LOW text using the regular fixed text layer, along with the scores. That solution would work on real hardware and is cleaner. The patch can live inside mame2mra.toml

Indeed, the gear function in software could even parse the extra DIP switch bit, so the hardware -the core- does not need to be modified. Because the toggle itself is a small thing in the RTL, I don't mind including that in hardware though. Whatever you prefer for that part.

niknakniknak:
Ok I've moved the gear indicator to a patch for most of the versions. outrunm is missing here so I can't patch that currently.  Toggle, I've kept in the core. that would be another patch per set so this seems more logical to me. Main difference visually is TOR LOW/HI is now yellow as white wasn't an in game palette option. otherwise same. only in game for both OR/TOR and doesn't show when TOR is in automatic.

jotego:
I simplified the TOML a bit.

Issue #1106, opened 2025-07-05
Outrun sprite glitches


niknakniknak:
Outrun has sprite glitches that I've noticed when you are far from the road. Noticeable on scenery and crashes occasionally. This shot is paused to catch it, but even paused it twitches and moves between the missing lines. I'll try and capture a video if needed.

![Image](https://github.com/user-attachments/assets/ca6a0065-8150-4a6d-b30f-4a961d1fb16e)

jotego:
Duplicated in #1035 #744 #227. This is a limitation of the core that would need a major re-work to get fixed. I may tackle this in the future as other work on other cores may bring up new ideas usable here too.

Issue #1035, opened 2025-04-11
outrun/toutrun: Glitchy graphics in Turbo Out Run


rp-jt:
At the start of Turbo Out Run there is a lot of glitched graphics which then disappear.
It can happen in game sometimes too

- Reported on Discord by Gills and Denymetanol

jotego:
This is a limitation of how the sprite frame buffer is implemented. I do not think we can improve it with the memories available on the MiSTer. There are no plans to change this.

Issue #744, opened 2024-07-28
Outrun graphic glitch on left side


theenemy-thd:
I have noticed an issue on the graphics on the left hand side especially on the stage with stone arches ... it happens to the left but also bad above on the top of the arch that goes the full length of the screen but again only on the left
![20240728_111820](https://github.com/user-attachments/assets/0b6b8b68-a0dc-425b-8333-84ae5ebbce65)

jotego:
Thanks for reporting it. This is due to a limitation in how many points can be drawn per line without using a frame buffer for the sprites. There is no fix planned.

theenemy-thd:
something that can be fixed on more powerful fpga?

jotego:
Particularly, one with more RAM chips

theenemy-thd:
You mean on board ram that has a higher access speed?

Issue #227, opened 2023-03-25
outrun: large objects may show a partially blank line on the left side


jotego:
Missing pixels at the top left of the screen only for some scenes with many sprites
![20230923_094606-screen](https://github.com/jotego/jtcores/assets/1863036/0355ee44-f9bd-42af-ae15-34a8b4a433f2)

Seen in 5b3830c6. 

- Not related to https://github.com/jotego/jtcores/blob/a4b5b39a2a3238168ef84ad4a49cfd28db0f2e87/cores/outrun/hdl/jtoutrun_obj_draw.v#L141

- Related to the line-based frame buffer?

jotego:
The blank lines appear at the same relative horizontal position in the object. Varying the horizontal zoom shows how the blanks follow the object and can appear closer to the border or the screen centre depending on the h-zoom setting.
The point at which the blanking turns to solid object is not fixed... bad data from the rom?

jotego:
- not related to the safe re-start in: https://github.com/jotego/jtcores/blob/a4b5b39a2a3238168ef84ad4a49cfd28db0f2e87/cores/outrun/hdl/jtoutrun_obj_draw.v#L148

jotego:
The problem occurs in the frame buffer. Some lines are not drawn because the request occurs too lat. Making `hlim` larger has a clear effect. `hblen`is $4F, by making:

`hlim   <= hcnt - (hblen-st_addr);`

It is possible to stabilize the lines gaps vertically setting `st_addr=$8`. The problem cannot be completely removed with this approach, though.

ksdomino:
This might not be a fixable bug. Outrun arcade can only " display 76 scenery sprites at any one time" so some arches are broken in the arcade version. source: https://reassembler.blogspot.com/2012/01/gateways-broken-arches.html

jotego:
Limiting it to 76 does not solve the issue as that limitation came from the software side of the system, not the hardware (the FPGA in our case). I think I'll settle with the current approach until an FPGA that enables a proper frame buffer shows up.

PR #1611, opened 2026-09-16
Outrun: Turbo outrun potential sprite glitch fix

niknakniknak:
A possible fix for the Turbo Out Run sprite glitches, loosely based on how Gals Panic handle their line buffer. Should cover #227, #744, #1035 and #1106.

From what was researched there are two things going on

jtoutrun_obj_scan signals ln_done when it has handed out the last sprite of the line, not when the drawer has actually finished drawing it. The frame buffer then starts copying the line to DDR while it's still being drawn, so anything drawn behind the copy gets lost. Run and Gun and Gals Panic already hold ln_done until their drawer is idle, so I've done the same in jtoutrun_obj.

It's still too slow on busy frames and the bottom lines don't get finished before VS. So I went a step further than those cores: the two line buffers swap as soon as ln_done arrives, and the finished line is copied to DDR while the next one is being drawn. The next line only starts once the swap has happened, so the drawer never writes into the buffer being copied.

frame188_before_vs_pipeline

Fully tested and played through on mister many times.

Opus 5.2 helped a lot with this one. I went through it and to me it looks okay. Up to you. thought I'd offer it.


jotego:
I made the drawing-wait in jtoutrun_obj the only path. That change makes sense for all versions. Sadly, that is not enough for the Pocket to show an improvement.

I am looking into expanding this to the other LF buffer versions. I will push more commits later.

asturur:
> I made the drawing-wait in jtoutrun_obj the only path. That change makes sense for all versions. Sadly, that is not enough for the Pocket to show an improvement.
> 
> I am looking into expanding this to the other LF buffer versions. I will push more commits later.

But the macro that changes the below generics in jtframe is mister only. I imagine you changed it locally before compiling when you tested for pocket. If that macro is really mister only because depends on something the other platforms do not have ( DDR or whatever ) can we give it a name that tells the story?

jotego:
I removed the macro altogether. It looks like it can be permanently used for MiSTer. In other platforms, it doesn't make sense. It only helps in MiSTer because it seems to work better with the DDR latencies.

For the other platforms, the frame buffer memory is not shared. I think the approach in this PR helps when there is contention in the frame buffer memory.

I am still surprised that the GFX problem completely goes away in MiSTer. I cannot test it myself right now.

In the Pocket, even doubling the frame buffer RAM speed is not enough to solve the problem.

niknakniknak:
yep its totally gone. TOR at the start is the best visual test for that game. for OR the best guaranteed visual test from what I've found is get past stage 1 and go to the right. on stage 2, full speed, smash into the rocks on the right side. it throws the car way far left and corrupts out on the current core.

As for pocket. don't have one so its never a priority for me. not surprised it wouldn't work for that though from what you've said.

jotego:
I need to test that it doesn't break the other games using the frame buffer before merging it

PR #1622, opened 2026-09-20
test(): Create regression rom for gng

asturur:
Custom test ROM for the GnG FPGA core with no copyrighted material. Boots with the real GnG hardware init sequence, displays a test screen with labeled input indicators and DIP switch glyphs.

Verified on MiSTer FPGA hardware.

Issue i encountered because i developed the rom on mame:
- FPGA char palette at $C0+ (pixel_mux prefix 2'b11), not $80 (MAME) -> the test was appearing in greys rather than color
- Palette writes must happen during VBLANK (colmix eff_AB mux) -> otherwise random palette every time
- DP register must be explicitly set to 0 after reset
- Boot must include sprite RAM fill + 100 SYNC cycles

Includes: assembly source, font generator, MRA, cab input script, Dockerfile for CI builds, and the pre-built regtest.zip.

It doesn't work on simulator yet

Plus i still need to undertand how to trigger it properly


cores/gng/reg/build.sh
```diff
@@ -0,0 +1,60 @@
...
+    MAME="${MAME:-$HOME/Emus/mame0276-arm64/mame}"
+    if [ ! -x "$MAME" ]; then
+        echo "ERROR: MAME not found at $MAME"
+        echo "Set MAME env var to your mame binary path"
+        exit 1
+    fi
+    echo "=== Launching MAME ==="
+    ln -sf regtest.zip "$BUILD_DIR/gng.zip"
+    "$MAME" gng -rompath "$BUILD_DIR" -window -nofilter -skip_gameinfo
+fi
```
asturur, line 60: we cannot have one of those per core. we need a different way of building and in go, not in python.
Still need work here

